PR #9, opened 2022-11-29
Make sdl2-compat even more compatible with SDL2

madebr:
At this speed, sdl2-compat is going to be more compatible to sdl2 that sdl2 itself :P

- Fixes SOVERSION + DYLIB compatibility/current ersioning
- Fixes libraries in sdl2_compat.pc
- Fixes test/testgles2_sdf.c
- Fixes error trying to include X11 when it is not available

PR #10, opened 2022-11-30
cmake: build .m files as objective-c

madebr:
Attempt to address https://github.com/libsdl-org/sdl2-compat/pull/9#issuecomment-1331528664

After this pr, `.m` sources should be explicitly compiled with `-x objective-c`.


sezero:
Let's rebase this to current master so that it goes through CI runs.

sezero:
Looking good. No issues in CI runs either. This is in.

PR #11, opened 2022-12-27
Add back SDL_FreeWAV

1bsyl:
Add back SDL_FreeWAV

but doesn't compile. 
because of SDL_oldnames : 
`#define SDL_FreeWAV SDL_FreeWAV_renamed_SDL_free`

not sure, if we should remove it from SDL3, or change in sdl2-compat

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -2772,6 +2772,12 @@ SDL_Has3DNow(void)
     return SDL_FALSE;
 }
 
+DECLSPEC void SDLCALL
+SDL_FreeWAV(Uint8 * audio_buf)
+{
+    return SDL_free(audio_buf);
+}
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -2775,7 +2775,7 @@ SDL_Has3DNow(void)
 DECLSPEC void SDLCALL
 SDL_FreeWAV(Uint8 * audio_buf)
 {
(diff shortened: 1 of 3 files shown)


madebr:
In SDL3, we should probably replace
https://github.com/libsdl-org/SDL/blob/04f2820e073053a15961ec6340a7b741b1c42de5/include/SDL3/SDL_oldnames.h#L365
with e.g.
```c
#elif !defined(SDL_OLD_NAMES_DO_NOTHING)
```
and do the following in sdl2-compat:
```c
#ifdef SDL_ENABLE_OLD_NAMES
#undef SDL_ENABLE_OLD_NAMES
#endif
#ifndef SDL_OLD_NAMES_DO_NOTHING
#define SDL_OLD_NAMES_DO_NOTHING
#endif
```

src/sdl2_compat.c
```diff
@@ -2772,6 +2772,12 @@ SDL_Has3DNow(void)
     return SDL_FALSE;
 }
 
+DECLSPEC void SDLCALL
+SDL_FreeWAV(Uint8 * audio_buf)
+{
+    return SDL_free(audio_buf);
```
sezero, line 2778: No `return` here: this is a `void` procedure.

1bsyl: done thanks!

1bsyl:
testing it, this looks ok to me

1bsyl:
(when building against a previous SDL3 revision..)

1bsyl:
merging, since this was tested!

PR #14, opened 2022-12-28
Add compat for Hints functions, because of hint name changes

1bsyl:
Add compat for Hints functions, because of hint name changes

SDL2 -> SDL3
SDL_AUDIODRIVER -> SDL_AUDIO_DRIVER
SDL_VIDEODRIVER -> SDL_VIDEO_DRIVER


1bsyl:
Merged since it was tested with env variable and calling the function.
only callback may need more tweaking

PR #15, opened 2023-01-04
Helps to compile and run

1bsyl:
not sure if this is 100% correct, but sdl2-compat tests programs  run now

- lock sensor / joystick (dummy stubs)
- add calls for render API  int


1bsyl:
merged since it runs testsprite2 etc

PR #18, opened 2023-01-04
Add Joystick / Sensor Locks

1bsyl:


1bsyl:
test

sezero:
Do we actually need this?

1bsyl:
we need the function sure, because there were in SDL2.
they could be empty maybe but not sure ...
but if an app make a deadlock in sdl2 we should do it we sdl2-compat :)

sezero:
I was leaning on the idea of keeping them empty. @slouken?

madebr:
My 2c is we need these implemented, because SDL2 promises stable results across device connects/disconnects.

> In particular, you are guaranteed that the sensor list won't change, so the API functions that take a sensor index will be valid, and sensor events will not be delivered.

PR #20, opened 2023-01-05
Add legacy audio function

1bsyl:
re-add legacy audio function that works with id=1
after this gets merged:  https://github.com/libsdl-org/SDL/pull/6999


icculus:
Nice work on this!

1bsyl:
thanks! was easier than expected :)

Issue #17, opened 2023-01-04
flicker / issue rendering ? 


1bsyl:
./testgeometry : triangle disappears when moving / click with the mouse. it should rotate the triangle instead.
./testsprite2 flickers (not sure ?)

not sure why:  mouse coords ? maybe missing sensor/ joysticks locks ? )

sezero:
Mouse coords became floats in SDL3, that should be the reason. Still a TODO:
https://github.com/libsdl-org/sdl2-compat/blob/main/src/sdl2_compat.c#L1357-L1358

1bsyl:
ok thanks ! 
maybe the issue, i haven't investigated yet

1bsyl:
maybe, does ./testgeometry compile ? if you run it ?

PR #21, opened 2023-01-05
Deal with mouse coords becoming floats in SDL3.

sezero:
With this, quakespasm seems to run stable with mouse motions, wheel and
buttons. Note the added FIXME in `SDL_MOUSEWHEEL` cases: Any correction
or changes?

Please review.


sezero:
@1bsyl: I guess this would be the cure for https://github.com/libsdl-org/sdl2-compat/issues/17

1bsyl:
maybe, does ./testgeometry compile ? if you run it ?

1bsyl:
(just need to enable BUILD_TEST in cmake)

1bsyl:
SDL2COMPAT_TESTS:BOOL=ON

icculus:
My plan was to change the SDL_MOUSEMOTION, etc events to different numbers in SDL3, so they don't overlap with SDL2, and generate new events in EventFilter3to2, but now I can't remember why I thought that was necessary, and this is probably the better way to do this if it _isn't_ necessary

sezero:
> maybe, does ./testgeometry compile ? if you run it ?

No

1bsyl:
indeed fixed!

slouken:
There was some accumulation of mouse deltas that was being done that will need to be replicated in sdl2-compat.

icculus:
I don't know what this is specifically either. :)

sezero:
> It's the code relating to GetScaledMouseDelta() in https://github.com/libsdl-org/SDL/blob/8c3239dee50f06a959b7e772570c716a757315ca/src/events/SDL_mouse.c

`git blame` shows `GetScaledMouseDelta` received updates during 2.26.0 development
as the latest (apart from SDL3 style changes.) Don't know what to do with this.

PR #27, opened 2023-01-16
Re-add SDL_AudioCVT interface, based on SDL_AudioStream

1bsyl:
If https://github.com/libsdl-org/SDL/pull/7092 gets merged ...


src/sdl2_compat.c
```diff
@@ -3914,6 +3914,231 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+    cvt->needed = 0;
+    cvt->filter_index = 0;
+    SDL_zeroa(cvt->filters);
+    cvt->len_mult = 1;
+    cvt->len_ratio = 1.0;
+    cvt->rate_incr = ((double)dst_rate) / ((double)src_rate);
+
+    /* Use the filters[] to store some data ... */
+    {
+        AudioParam *ptr = (AudioParam *)&cvt->filters[1];
```
icculus, line 4024: Let's move this to the very last possible position in the array, so there's less chance that an app will overwrite it if it tries to do something dangerous like add its own functions to the end of the filter list.

(SDL2 did this, too, in SDL_ResampleCVT, for the same reason.)

src/sdl2_compat.c
```diff
@@ -3914,6 +3914,231 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+
+        ptr->src_format = src_format;
+        ptr->src_channels = src_channels;
+        ptr->src_rate = src_rate;
+        ptr->dst_format = dst_format;
+        ptr->dst_channels = dst_channels;
+        ptr->dst_rate = dst_rate;
+
+        cvt->needed = 1;
+        if (src_format == dst_rate && src_rate == dst_rate && src_channels == dst_channels) {
```
icculus, line 4035: `src_format == dst_rate` is a copy/paste error.

src/sdl2_compat.c
```diff
@@ -3914,6 +3914,231 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+        }
+       
+        if (src_channels < dst_channels) {
+            cvt->len_mult = ((cvt->len_mult * dst_channels) + (src_channels - 1)) / src_channels;
+        }
+
+        /* !!! FIXME: remove this if we can get the resampler to work in-place again. */
+        /* the buffer is big enough to hold the destination now, but
+           we need it large enough to hold a separate scratch buffer. */
+        cvt->len_mult *= 2;
```
icculus, line 4055: Since this is going to push through an SDL_AudioStream that manages its own buffer, we don't need this anymore, since we don't need the app to allocate twice as much space just to make sure we have enough scratch memory to do the work in-place in their buffer.

src/sdl2_compat.c
```diff
@@ -3914,6 +3914,231 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+    }
+    if (src_rate >= SDL_MAX_SINT32 / RESAMPLER_SAMPLES_PER_ZERO_CROSSING) {
+        return SDL_SetError("Source rate is too high");
+    }
+    if (dst_rate >= SDL_MAX_SINT32 / RESAMPLER_SAMPLES_PER_ZERO_CROSSING) {
+        return SDL_SetError("Destination rate is too high");
+    }
+#endif
+
+    stream = SDL3_CreateAudioStream(src_format, src_channels, src_rate,
```
icculus, line 4001: We're creating a stream here that we don't use.

src/sdl2_compat.c
```diff
@@ -3914,6 +3914,231 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+
+    src_len = cvt->len & ~(src_samplesize - 1);
+    dst_len = dst_samplesize * (src_len / src_samplesize);
+    if (src_rate < dst_rate) {
+        const double mult = ((double)dst_rate) / ((double)src_rate);
+        dst_len *= (int) SDL_ceil(mult);
+    }
+
+    dst_len = dst_len & ~(dst_samplesize - 1);
+    dst_buf = (Uint8 *)SDL_malloc(dst_len);
```
icculus, line 4110: So the AudioCVT struct tells the app how big its buffer needs to be for not just the result, but also for any work the converter needs to do. We shouldn't need to malloc an output buffer here just to memcpy to the actual output buffer, as the app will provide us a buffer as large as we request and we can write directly into it.

icculus (changes requested):
This is good, just needs a few changes. This API has been confusing people since SDL 1.0, so it's only fitting it bite us one more time.  :)

src/sdl2_compat.c
```diff
@@ -1111,7 +1111,7 @@ SDL_GetRevisionNumber(void)
 }
 
 
-DECLSPEC void SDLCALL
+DECLSPEC int SDLCALL
```
sezero, line 1114: This, and the corresponding change to SDL_OpenAudio below, can go in directly without this PR. (How did we miss this?..)

src/sdl2_compat.c
```diff
@@ -2986,8 +2987,7 @@ SDL_OpenAudio(SDL_AudioSpec *desired, SDL_AudioSpec *obtained)
     }
 
     if (g_audio_id > 0) {
-        SDL_SetError("Audio device is already opened");
-        return -1;
+        return SDL_SetError("Audio device is already opened");
```
sezero, line 2990: .. and why aren't we directly calling `SDL3_SetError` here?

src/sdl2_compat.c
1bsyl: yes done!

src/sdl2_compat.c
1bsyl: yes removed, (wasn't sure if I could )

src/sdl2_compat.c
1bsyl: I thought this would be useful as a sanity check.
should I remove it anyway ?

src/sdl2_compat.c
1bsyl: here, I have some trouble :/
are you sure I can put 
`SDL3_PutAudioStreamData(stream, cvt->buf, src_len)`
then read on the same buffer ?
`real_dst_len = SDL3_GetAudioStreamData(stream, cvt->buf, dst_len) `

because when I tried with `./testresample`, if make a different output wav with has some crackling noise

src/sdl2_compat.c
1bsyl: yes, don't know :) at first I though this was intended but let's fix it within this pr anyway

src/sdl2_compat.c
1bsyl: yes updated!

src/sdl2_compat.c
1bsyl: yes typo using completion:)

src/sdl2_compat.c
1bsyl: ok. done there was a mistake in my previous code. memcpy was a wrong size.

src/sdl2_compat.c
```diff
@@ -3914,6 +3914,229 @@ DECLSPEC void SDLCALL SDL_SIMDFree(void *ptr)
...
+{
+    /* Sanity check target pointer */
+    if (cvt == NULL) {
+        return SDL_InvalidParamError("cvt");
+    }
+
+    /* Make sure we zero out the audio conversion before error checking */
+    SDL_zerop(cvt);
+
+#if 0
```
sezero, line 3972: Should we `#if 0` the `SDL_IsSupportedAudioFormat` and `SDL_IsSupportedChannelCount` procedures too?

src/sdl2_compat.c
icculus: > (How did we miss this?..)

Likely it was in the initial framework I copy/pasted from sdl12-compat, where this didn't return an int.

src/sdl2_compat.c
icculus: It allocates memory and does a lot of sanity checks that we're doing here anyhow, so I'm inclined to say it should be removed.

icculus:
(Also, thanks to @1bsyl for the patience, as we're really beating up on this PR.)

src/sdl2_compat.c
1bsyl: ok, 
so: pre-creating audio stream is removed.
and re-added the checks

src/sdl2_compat.c
1bsyl: now this is re-enabled !

1bsyl:
> (Also, thanks to @1bsyl for the patience, as we're really beating up on this PR.)
thnaks for your patience too :)
git/github makes this back and forth modifications quite easy !

1bsyl:
merging, since all was resolved

PR #32, opened 2023-01-29
Update DisplayMode awareness

1bsyl:
@slouken 
updated, please double-check!


1bsyl:
commiting this was based on slouken comment: 
https://github.com/libsdl-org/sdl2-compat/commit/b4b4c95d866fb4fae5b7c0245d568be68db900c1#comments


>  this should be screen_w/h
> sdl2-compat should set the hint "SDL_WINDOWS_DPI_SCALING" to "0" by default.
> It should also set the hint "SDL_WINDOWS_DPI_AWARENESS" to "unaware" by default.

PR #35, opened 2023-01-31
Adjust for new SDL_WINDOW_FULLSCREEN behavior

sezero:
After https://github.com/libsdl-org/SDL/pull/7174.

@slouken: is this accurate?


slouken:
Yep

sezero:
OK, pulled it in.

PR #41, opened 2023-02-02
Fix build with fullscreen-desktop

1bsyl:


1bsyl:
oops, I've run with wrong libs. it didn't load the correct lib. so this builds, but still doesn't run

PR #48, opened 2023-02-03
Joysticks: change instance id to index in events

1bsyl:


1bsyl:
here, the PR modify the event in-place, replacing instanceId with an index.

If re-pushed an SDL2 event, it crashes (maybe it goes in loop and overflow). Even If I returns 0 from the event filter not to add the SDL3 event.

1bsyl:
comiting. so this at least can be tested since the library evolves quickly

PR #66, opened 2023-03-09
Add shaped window based on SDL_WINDOW_TRANSPARENT

1bsyl:
If Shaped Window get removed in SDL3, 
this add back the functionality in SDL2-compat, but only for SDL_Renderer and for 1 window (which could be extended).


1bsyl:
(testshape doesn't work perfectly, because there is one bug outside the scope)

sezero:
@1bsyl, @slouken: I rebased this so that it applies cleanly. (Build-tested only, no run-testing!)

Either this, or some new implementation is needed after shaped window removal in SDL3.

slouken:
I have this covered, thanks!

sezero:
>  but only for SDL_Renderer and for 1 window (which could be extended).

Can these limitations be lifted?

> I have this covered, thanks!

Something changed?

slouken:
Yes, the limitations can be lifted. I was working on something that did that, but I ended up deciding it was better to merge than leave it broken for a few days.

PR #72, opened 2023-04-20
Convert sensor timestamp

meyraud705:
Sensor timestamps are in nanosecond in SDL3, microsecond in SDL2.


sezero:
This looks OK to me. @slouken ?

sezero:
OK then, this is in.

PR #73, opened 2023-05-02
Add basic popup support

Kontrabant:
Adds basic support for popup windows in SDL2 applications.

The implementation is based on the SDL2 Wayland implementation of popups originally by flibitijibibo, as the cross-platform popup implementation in SDL3 behaves in much the same way with popups being positioned relative to the parent. New popup windows are spawned as children of the window with mouse focus, and tooltips follow the cursor. This is known to be sufficient for Unreal Engine's use of them.

Are there any other known cases of applications using the tooltip or popup menu types with SDL2? If anything was specifically targeting X11 and expects to get a raw window with these flags to be passed through, there may be an issue, but from a cursory search on Github and Google, I can't really find anything. These flags were added for Unreal to begin with and are still tagged 'X11 only' in the SDL2 documentation, so it's doubtful that much else uses them.

If this basic "as required by Unreal" implementation turns out to be insufficient for some purpose, it can always be expanded upon.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -4593,11 +4593,14 @@ SDL_GetWindowFlags(SDL_Window *window)
     return flags;
 }
 
+#define POPUP_PARENT_PROP_STR "__SDL3_parentWnd"
+
 DECLSPEC SDL_Window * SDLCALL
 SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
 {
-    SDL_Window *window;
+    SDL_Window *window = NULL;
     int hidden = flags & SDL_WINDOW_HIDDEN;
+    const Uint32 is_popup = flags & (SDL_WINDOW_POPUP_MENU | SDL_WINDOW_TOOLTIP);
 
     flags &= ~SDL2_WINDOW_SHOWN;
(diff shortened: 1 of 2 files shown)


slouken:
The tooltip following mouse behavior seems like it should be implemented in the application. This definitely isn't something every application wants.

Steam uses them, FYI.

Kontrabant:
Reworked things, and popup positioning is now always explicit.  The UE5 editor seems to work in near identical fashion running under sdl2-compat as it does under native SDL2, which is to say that on X11 it's fine, while Wayland runs, but has bugs (mouse input being the big one, as it wants global coordinates in certain cases, which Wayland can't provide).

This shook out a few bugs in the SDL3 code too.

> Steam uses them, FYI.

Isn't Steam already on SDL3, so sdl2-compat behavior shouldn't affect it?

slouken:
Yep, this looks good to me. :)

Thanks!

Kontrabant:
One more fix, as moving the cursor quickly between menus in the UE5 editor could cause mouse focus to be briefly lost while at the same time trying to create a new popup window, which would fail due to the mouse focus retrieval function returning null, which would crash the editor.  Falling back to keyboard focus if the mouse focus returns null fixes this.

PR #75, opened 2023-05-19
SDL_ConvertAudio: Fix source/destination format confusion 

sulix:
The SDL_ConvertAudio() function accidentally uses the destination format as the source format, breaking (presumably amongst other things) 8-bit WAV playback in Uplink under sdl12-compat.

Note that this change is currently based on top of #74, just so it compiles. Only the second commit (with the s/dst_format/src_format) actually matters here.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3655,12 +3655,10 @@ static void
 DisplayMode_2to3(const SDL2_DisplayMode *in, SDL_DisplayMode *out) {
     if (in && out) {
         out->format = in->format;
-        out->pixel_w = in->w;
-        out->pixel_h = in->h;
-        out->screen_w = in->w;
-        out->screen_h = in->h;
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -3798,9 +3796,9 @@ SDL_GetDisplayDPI(int displayIndex, float *ddpi, float *hdpi, float *vdpi)
     }
 
 #if defined(__ANDROID__) || defined(__IOS__)
-    val = dp->display_scale * 160.0f;
+    val = dp->pixel_density * 160.0f;
 #else
-    val = dp->display_scale * 96.0f;
+    val = dp->pixel_density * 96.0f;
```
slouken, line 3801: This should actually be the content scale, not the pixel density.

slouken: Fixed in ca1c0aa

PR #78, opened 2023-05-29
Fix joystick events index/instance ID

meyraud705:
This patch fix 2 issues with joystick event:

 - Only "ADDED" events use joystick index.

 - Translate event as late as possible so we don't need to update "ADDED" event already in the queue when a device is removed, this was done in `UpdateEventsForDeviceRemoval()` in SDL2. I also think it is not a good idea to modify SDL3 events while they are in SDL3 queue.


slouken:
This looks great, thanks!

PR #80, opened 2023-06-27
audio: Updated for the SDL3 audio redesign.

icculus:
This PR updates sdl2-compat to use the new SDL3 audio API, which is currently sitting in https://github.com/libsdl-org/SDL/pull/7704

This is a draft pull request because the SDL pull request has not yet been merged. Once it's merged, this should be pulled in, too.

My extremely-limited testing shows this works pretty well. The audio tests: loopwave, loopwavequeue, testaudioinfo, testmultiaudio, testsurround, and testaudiocapture...these all appear to work.

The only thing failing at the moment is testhotplug.c, because the audio events do not have the correct mappings from SDL3 instance IDs to SDL2 indexes. We need to catch these, add new audio devices to the list, and send the event on to SDL2 land with the correct values. Device removal needs to map to open device IDs and send a device id of zero for removal of physical hardware that isn't opened. We don't have to remove from the list of devices, though, since this is rebuilt during calls to SDL_GetNumAudioDevices() in SDL2.

Anyhow, that's still pending, but this is solid enough that it can go in revision control as a first pass and deal with that issue in a later commit.


sezero:
Here is a rebased-to-current patch build-tested against your latest audio-redesing tree: [patch.txt](https://github.com/libsdl-org/sdl2-compat/files/12177999/patch.txt)

icculus:
Force-pushed this to the latest, with @sezero's updates. Clicking "merge" on this one after the GitHub Actions verify it.

PR #81, opened 2023-07-03
Use libsdl-org/setup-sdl GitHub action

madebr:
Looks like testautomation and testfile fail, on ci as well as on my machine.

Also, could anybody with an Apple look at the Macos timeouts?
The `libsdl-org/setup-sdl` action sets `DYLD_LIBRARY_PATH`.


slouken:
I'm working on getting the SDL3 tests passing and then I'll take a look at the SDL2 tests, and can check it out on macOS

madebr:
Thanks @sezero . The `testfile` test succeeds now.
However, in the mean time linux has regressed a bit: there is now a segmentation fault somewhere in the testautomation clipboard test.

Currently, by default, the libsdl-org/setup-sdl action does not install any dependencies, so there is no X11 backend.
Perhaps this is causing issues? (Should the default be changed?)

sezero:
Don't know how setup-sdl works at all. However, we should run the testsuite selectively if possible: even `keyboard_getKeyFromScancode` of testautomation fails for me because of my keyboard layout. There also audio resample test failures among others.

slouken:
@icculus and I can revisit the automated tests soon. Are these failing on main or sdl2-compat?

madebr:
I think SDL3 main has issues. See #7466 for that.

madebr:
To address https://github.com/libsdl-org/SDL/issues/8040, the sdl2-compat workflow need something as follows:
```
on:
  schedule:
    - cron: '5 1 * * *' # Build every day at 1:05am
```

.github/workflows/main.yml
```diff
@@ -7,31 +7,52 @@ jobs:
...
+        sudo apt-get install -y libgl1-mesa-dev libglu1-mesa-dev
+    - name: Configure (CMake)
+      shell: sh
+      run: |
+        cmake -B build -GNinja
+    - name: Build (CMake)
+      run: |
+        cmake --build build/
+    - name: Run build-time tests (CMake)
+      if: ${{ false }} # FIXME: enable build-time tests on CI
```
madebr, line 47: I left running the self tests disabled.
A follow-up PR will re-enable this line.

madebr:
> To address [libsdl-org/SDL#8040](https://github.com/libsdl-org/SDL/issues/8040), the sdl2-compat workflow needs something as follows:
> 
> ```
> on:
>   schedule:
>     - cron: '5 1 * * *' # Build every day at 1:05am
> ```


Do we want this?

slouken:
*deep breath*

Sure.

PR #82, opened 2023-07-11
fix RWops read and write return values.

sezero:
`testfile` program fails without this. Tested on i686-linux.


slouken (approved):
I didn't realize the SDL2 APIs returned number of elements, but this looks fine to me.

sezero:
@madebr: Rebase your setup-sdl patch for one less testsuite failure

icculus:
> I didn't realize the SDL2 APIs returned number of elements

Yeah, inspired by fwrite(), not write(). I wasn't sad to change this for SDL3.

PR #87, opened 2023-07-27
override SDL_Quit and SDL_QuitSubsystem, adjust the gesture code.

sezero:
@icculus, @slouken: Please review this carefully.


src/sdl2_compat.c
```diff
@@ -2559,37 +2559,11 @@ GestureAddTouch(const SDL_TouchID touchId)
...
-    return 0;
-}
-
 static GestureTouch *
 GestureGetTouch(const SDL_TouchID touchId)
 {
-    int i;
-    for (i = 0; i < GestureNumTouches; i++) {
+    int i = 0;
+    for (; i < GestureNumTouches; i++) {
```
slouken, line 2566: Why would you make this change? It's typical for loop variables to be initialized as part of the loop construct.

src/sdl2_compat.c
```diff
@@ -2559,37 +2559,11 @@ GestureAddTouch(const SDL_TouchID touchId)
     return &GestureTouches[GestureNumTouches++];
 }
 
-static int
-GestureDelTouch(const SDL_TouchID touchId)
```
slouken, line 2563: When was this function called in SDL2? I thought this was dynamically removing touches?

src/sdl2_compat.c
sezero: I usually do it that way if it's done only once in a procedure like this one. Can revert that part if you want.

src/sdl2_compat.c
sezero: Was called from within `SDL_DelTouch` in SDL_touch.c. Nowhere to call in here.

src/sdl2_compat.c
slouken: Yes please.

slouken: This looks good to me, once you've reverted this change.

sezero: Done

PR #88, opened 2023-07-29
Don't pass removed flags to SDL_CreateRenderer()

meyraud705:
`SDL_RENDERER_TARGETTEXTURE` was removed in https://github.com/libsdl-org/SDL/commit/dcd17f547324a143d66d79e3b586577a7da558ed

Trying to pass `SDL_RENDERER_TARGETTEXTURE`  to `SDL_CreateRenderer()` result in the following error:  `Couldn't find matching render driver`.

This patch clear the flag before passing it to SDL3. From `SDL_GetRenderDriverInfo()`, I assume this flag is supported by all render, so it is safe to ignore it.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3239,15 +3239,25 @@ SDL_GetRenderDriverInfo(int index, SDL_RendererInfo *info)
 DECLSPEC SDL_Renderer *SDLCALL
 SDL_CreateRenderer(SDL_Window *window, int index, Uint32 flags)
 {
+    SDL_Renderer *renderer = NULL;
     const char *name = NULL;
+    int want_targettexture = flags & SDL2_RENDERER_TARGETTEXTURE;
     if (index != -1) {
         name = SDL3_GetRenderDriver(index);
         if (!name) {
             return NULL;  /* assume SDL3_GetRenderDriver set the SDL error. */
         }
     }
+
     flags = flags & ~SDL2_RENDERER_TARGETTEXTURE; /* clear flags removed in SDL3 */
-    return SDL3_CreateRenderer(window, name, flags);
+    renderer = SDL3_CreateRenderer(window, name, flags);
+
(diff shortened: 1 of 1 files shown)


sezero:
SDL_SetRenderTarget documentation in SDL_render.h still mentions SDL_RENDERER_TARGETTEXTURE though, which needs correcting:
https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_render.h#L781-L800

Besides, in SDL_RenderTargetSupported(), we check the result for SDL2_RENDERER_TARGETTEXTURE: What to do about that?
https://github.com/libsdl-org/sdl2-compat/blob/main/src/sdl2_compat.c#L3252-L3264

meyraud705:
I was wrong, "opengles" driver does not support `SDL_RENDERER_TARGETTEXTURE` (from SDL_GetRenderDriverInfo()). That's why SDL_RenderTargetSupported() checks it. Let me add a check in SDL_CreateRenderer() too.

meyraud705:
I added the flag check in `SDL_CreateRenderer()` and fixed `SDL_RenderTargetSupported()`. It checks support of render target by name like `SDL_GetRenderDriverInfo()` does.

I don't know why this flag was removed so someone else should fix the documentation of `SDL_SetRenderTarget()` and the migration readme.

sezero:
This looks OK to me. @slouken?

src/sdl2_compat.c
```diff
@@ -3239,15 +3239,25 @@ SDL_GetRenderDriverInfo(int index, SDL_RendererInfo *info)
 DECLSPEC SDL_Renderer *SDLCALL
 SDL_CreateRenderer(SDL_Window *window, int index, Uint32 flags)
 {
+    SDL_Renderer *renderer = NULL;
     const char *name = NULL;
+    int want_tartgettexture = flags & SDL2_RENDERER_TARGETTEXTURE;
```
slouken, line 3244: `want_targettexture`?

meyraud705: Took me some time to spot the typo, but it is fixed now.

PR #90, opened 2023-07-30
Don't forward SDL_GetRevision

### LINKED ISSUE #89: SDL_GetRevision broken
We currently do a pass-through with SDL_GetRevision which is wrong:
it reports SDL3 version and revision: We need something specific to
sdl2-compat.

```sh
$ LD_LIBRARY_PATH=~/sdl3/lib ./testver
INFO: Compiled with SDL 2.0 or newer
INFO: Compiled version: 2.29.0 ()
INFO: Linked version: 2.90.0 (SDL-prerelease-3.0.0-2044-gb221b59)
```

Also note the empty revision string in `Compiled version` case which
directly references `SDL_REVISION`: It looks like it cannot find the
generated SDL_revision.h, i.e. it prioritizes top-level include/SDL2
instead of build-dir. CC: @madebr

madebr:
~The include paths were added as `INTERFACE`, which means they are only visible to consumers of SDL2 (e.g. the test programs).
By making them `PUBLIC`, they are visible to all.~
Making them `PUBLIC` turned out to be not necessary: use a "new" `SDL2COMPAT_REVISION` instead.

Fixes #89


src/sdl2_compat.c
```diff
@@ -24,6 +24,7 @@
 #include "sdl3_include_wrapper.h"
 #include "sdl2_compat.h"
 #include "dynapi/SDL_dynapi.h"
+#include "SDL2/SDL_revision.h"
```
madebr, line 27: I'm unsure whether this is desired.

sezero: It might, but ugh.

Can we not add it to CFLAGS as a new SDL2COMPAT_REVISION define (private to library only) and return that instead of SDL_REVISION?

madebr: Force pushed a fix.
When no `SDL2COMPAT_REVISION` is passed, it will return `"SDL-2.90.0-unknown"`.

sezero: Looks good to me.

If have to be picky: Maybe `-no-vcs` instead of `-unknown` to follow our showrev.sh?? I'm fine either way, though.

slouken:
I'm thinking we follow the same convention for sdl2-compat that we did for sdl12-compat - we use the same minor version as SDL 2 (2.28) and bump the patchlevel to 50.

If we hit 2.28.48, we'll automatically roll over to 2.30.0 and release a new sdl2-compat 2.30.50.

@icculus, thoughts?

build-scripts/test-versioning.sh
```diff
@@ -0,0 +1,106 @@
...
+    not_ok "sdl2_compat.c SDL2_COMPAT_VERSION_PATCH $micro disagrees with SDL_version.h $ref_micro"
+fi
+
+ref_dylib_compatibility_version="$((100*$ref_minor + $ref_micro + 1)).0"
+ref_dylib_current_version="$((100*$ref_minor + $ref_micro + 1)).0"
+ref_dylib_versions="$ref_dylib_compatibility_version $ref_dylib_current_version"
+dylib_versions=$(sed -Ene 's/^LDFLAGS\+= -Wl,-compatibility_version,([0-9.]+) -Wl,-current_version,([0-9.]+)$/\1 \2/p' src/Makefile.darwin)
+
+if [ "$ref_dylib_versions" = "$dylib_versions" ]; then
+    ok "Makefile.darwin LDFLAGS $dylib_versions"
```
sezero, line 100: Can we do something similar for Makefile.linux (`SHLIB` in there)?

madebr: Now we do :)

sezero: Thanks!

sezero (approved):
This is OK by me.

I guess we're waiting for the answer to @slouken's versioning question?
(Me, I'm happy with 2.90.x with x increasing to an even number every release.)

src/sdl2_compat.c
```diff
@@ -1172,6 +1179,11 @@ SDL_GetRevisionNumber(void)
     return 0;
 }
 
+DECLSPEC const char * SDLCALL
+SDL_GetRevision(void)
+{
+    return SDL2COMPAT_REVISION;
```
madebr, line 1185: Just asking.
Is there a need to include SDL3's `SDL_GetRevision` in the returned string?
So this function returns a concatenation of SDL2-compat's revision and SDL3's `SDL_GetRevision`.
I suppose this would be useful for bug reports.

sezero: With `SDL2COMPAT_DEBUG_LOGGING` env var set, we hit the following in `LoadSDL3()`:
```
SDL3_Log("sdl2-compat 2.%d.%d, talking to SDL3 %d.%d.%d", SDL2_COMPAT_VERSION_MINOR, SDL2_COMPAT_VERSION_PATCH, v.major, v.minor, v.patch);
```
We can change that to report result of SDL3_GetRevision() maybe if really needed?

madebr: I was just asking. Thanks for reminding me about `SDL2COMPAT_DEBUG_LOGGING`.

slouken:
> This is OK by me.
> 
> I guess we're waiting for the answer to @slouken's versioning question? (Me, I'm happy with 2.90.x with x increasing to an even number every release.)

If we end up adding any new APIs, applications checking the minor version will think that older sdl2-compat versions will have them, so we should definitely not do that.

sezero:
> > This is OK by me.
> > I guess we're waiting for the answer to @slouken's versioning question? (Me, I'm happy with 2.90.x with x increasing to an even number every release.)
> 
> If we end up adding any new APIs, applications checking the minor version will think that older sdl2-compat versions will have them, so we should definitely not do that.

We can increment 90, then? (Just a naive thought.)

sezero:
PING ?

slouken:
> > > This is OK by me.
> > > I guess we're waiting for the answer to @slouken's versioning question? (Me, I'm happy with 2.90.x with x increasing to an even number every release.)
> > 
> > 
> > If we end up adding any new APIs, applications checking the minor version will think that older sdl2-compat versions will have them, so we should definitely not do that.
> 
> We can increment 90, then? (Just a naive thought.)

No, if sdl2-compat starts at 2.90, and we add an API in 2.30, and then add it in sdl2-compat at 2.92, then applications linked with the original 2.90 will think it's available. sdl2-compat should always match the major and minor version of SDL2 that it mimics. 2.28.50 was an arbitrary point, maybe it makes more sense to start at 100.

Let's do that. sdl2-compat will be 2.x.100+, where x is the minor version of SDL2 that it emulates.

All this might become moot if SDL2 never revs meaningfully, but it would make me feel better to start at a point that allows some flexibility in the future.

slouken:
@smcv might also have some feedback on the versioning for sdl2-compat that makes sense in the context of distribution releases.

smcv:
> No, if sdl2-compat starts at 2.90, and we add an API in 2.30, and then add it in sdl2-compat at 2.92, then applications linked with the original 2.90 will think it's available.

Yes, this. If we want the freedom to add new APIs/ABIs to "real SDL2" then we need sdl2-compat to mimic its major.minor version like you said.

Its on-disk name also needs to mimic SDL2 if it doesn't already, so sdl2-compat version 2.28.50 would have to be `libSDL2-2.0.so.0.2800.50` or something.

> sdl2-compat will be 2.x.100+, where x is the minor version of SDL2 that it emulates

Careful: do we have to limit version numbers to 2 digits for compatibility with versioning macros? If yes, then we should start at a smaller boundary, like 50 (or even 20, I don't think our micro versions have ever gone into 2 digits).

smcv:
> Careful: do we have to limit version numbers to 2 digits for compatibility with versioning macros?

Yes we do, the `SDL_VERSIONNUM` macro assumes that *z* fits in 2 digits.

smcv:
> If we hit 2.28.48 [in "real" SDL2], we'll automatically roll over to 2.30.0 [in "real" SDL2] and release a new sdl2-compat 2.30.50 [to match it]

That makes sense to me.

sezero:
OK then: When @madebr changes the version to 2.28.50 this is good to go, yes?

smcv:
To avoid this sort of thing all over again for SDL4, it would be good to make sure that either SDL3 doesn't have any interfaces that arbitrarily limit the width of a version number, or if it has those interfaces (for example limiting each component to 255 like SDL2's `SDL_version` does) then we're clear about documenting what the rules are.

src/Makefile.linux
```diff
@@ -14,7 +14,7 @@ LDFLAGS = -shared -Wl,-soname,libSDL2-2.0.so.0
 LDFLAGS+= -Wl,--no-undefined
 LDLIBS  = -ldl
 
-SHLIB = libSDL2-2.0.so.0.2900.0
+SHLIB = libSDL2-2.0.so.0.2800.0
```
slouken, line 17: I don't think this is right. What is the output filename when built with CMake?

smcv: I would expect that for 2.28.50, it should be `libSDL2-2.0.so.0.2800.50`.

smcv: Certainly with my packager hat on, I'm always going to prefer a conventional build system (Meson, CMake, Autotools) over a hand-crafted Makefile. I haven't packaged sdl2-compat for Debian experimental yet, but I used CMake for sdl12-compat and SDL3.

madebr: This is fixed now. I forgot how complex the numbering scheme was and pushed a bit too early.

I think these files are inherited from sdl12-compat.
I'll remove them.

icculus: Makefile.* is Ozkan's thing, I don't think it's meant for use by the general public.

sezero: > Yeah, I'm thinking we don't need Makefile.linux or Makefile.darwin. Why were these added?

So that I won't have to deal with cmake crapola everytime. They make life easier for me.

sezero:
This is ready to merge I guess?

P.S.: Maybe `update-version.sh` need updating some time.

madebr:
> This is ready to merge I guess?
> 
> P.S.: Maybe `update-version.sh` need updating some time.

I didn't see that script. It lives now at `build-scripts/update-version.sh`.

sezero:
OK, let's get this in.

PR #93, opened 2023-08-06
ci: test x86/x64/arm32/arm64 MSVC

### LINKED ISSUE #91: MSVC ARM32 builds fail linkage
Because we link with `/NODEFAULTLIB` :

```
sdl2_compat.obj : error LNK2019: unresolved external symbol __rt_sdiv referenced in function SDL_BuildAudioCVT_REAL
sdl2_compat.obj : error LNK2019: unresolved external symbol __rt_udiv referenced in function SDL_MixAudioFormat_REAL
sdl2_compat.obj : error LNK2019: unresolved external symbol __rt_udiv64 referenced in function EventFilter3to2
sdl2_compat.obj : error LNK2019: unresolved external symbol memcpy referenced in function SDL_JoystickGetDeviceGUID_REAL
```

### LINKED ISSUE #92: add MSVC arm64 and mingw32/64 runs to CI
@madebr: can you help with that?

P.S.: MSVC ARM32 currently fails linkage (see #91)

madebr:
Fixes #91
Fixes #92


.github/workflows/main.yml
```diff
@@ -4,27 +4,50 @@ on: [push, pull_request]
...
         platform:
-        - { name: Windows, os: windows-latest, msvc: true }
-        - { name: Linux,   os: ubuntu-latest }
-        - { name: MacOS,   os: macos-latest }
+        - { name: 'Linux',                    os: ubuntu-latest,  shell: sh }
+        - { name: 'MacOS',                    os: macos-latest,   shell: sh }
+        - { name: 'Windows msys2 (mingw32)',  os: windows-latest, shell: 'msys2 {0}', msystem: mingw32, msys-env: mingw-w64-i686 }
+        - { name: 'Windows msys2 (mingw64)',  os: windows-latest, shell: 'msys2 {0}', msystem: mingw64, msys-env: mingw-w64-x86_64 }
+        - { name: 'Windows msys2 (clang32)',  os: windows-latest, shell: 'msys2 {0}', msystem: clang32, msys-env: mingw-w64-clang-i686 }
+        - { name: 'Windows msys2 (clang64)',  os: windows-latest, shell: 'msys2 {0}', msystem: clang64, msys-env: mingw-w64-clang-x86_64 }
```
madebr, line 18: I don't understand why clang64 is failing during CMake configuration.
It fails in libsdl-org/setup-sdl, but also [during normal CMake configuration](https://github.com/madebr/sdl2-compat/actions/runs/5777560250/job/15657876918).

sezero: Don't know about that one..

The clang32 one does configure, but fails linkage because it can't do `-lgcc`: maybe adjust stuff so it links `compiler_rt` or whatever the clang version is supposed to be?

madebr: Looks like I couldn't do `path-type: inherit` for the `msys2/setup-msys2` action, when using clang64.

.github/workflows/main.yml
```diff
@@ -4,27 +4,49 @@ on: [push, pull_request]
...
+        - { name: 'Windows msys2 (ucrt64)',   os: windows-latest, shell: 'msys2 {0}', msystem: ucrt64,  msys-env: mingw-w64-ucrt-x86_64 }
```
sezero, line 19: Do we really need the ucrt version? We already build with no libc/crt dependency when targeting windows.

madebr: msys2 advises to use ucrt64 (https://www.msys2.org/docs/environments/)

madebr: But feel free to prune combinations. We don't want ci to become excessive.
Or select a few for ci, and run everything in a cron job.

CMakeLists.txt
```diff
@@ -278,7 +278,9 @@ if(MSVC)
   if(SDL_CPU_X86)  # don't emit SSE2 in x86 builds
     target_compile_options(SDL2 PRIVATE /arch:SSE)
   endif()
-  set_target_properties(SDL2 PROPERTIES LINK_FLAGS "/NODEFAULTLIB")
+  if(NOT MSVC_CLANG AND NOT SDL_CPU_ARM32)
```
sezero, line 281: Is `MSVC_CLANG` not `clang-cl`? I think we should simply check `MSVC` here??

.github/workflows/main.yml
sezero: My suggestion is dropping ucrt here.

cmake/sdlcompilers.cmake
```diff
@@ -0,0 +1,144 @@
...
+  set(USE_GCC FALSE)
+  set(USE_INTELCC FALSE)
+  set(USE_QCC FALSE)
+  if(CMAKE_C_COMPILER_ID MATCHES "Clang|IntelLLVM")
+    set(USE_CLANG TRUE)
+    # Visual Studio 2019 v16.2 added support for Clang/LLVM.
+    # Check if a Visual Studio project is being generated with the Clang toolset.
+    if(MSVC)
+      set(MSVC_CLANG TRUE)
+    endif()
```
madebr, line 12: This is the condition for `MSVC_CLANG`.

sezero: That is indeed clang-cl.

CMakeLists.txt
```diff
@@ -263,16 +269,17 @@ else()
 endif()
 
 if(MINGW)
-  target_link_libraries(SDL2 PRIVATE -nostdlib)
```
sezero, line 266: Why are we removing this?

.github/workflows/main.yml
madebr: I'll remove it once everything's in the green.

sezero:
We don't have to go overboard with -Wundef and documentation warnings: They are directly inherited from SDL2.

CMakeLists.txt
madebr: I'm trying things out, since I think I saw an error about missing libgcc somewhre.
And SDL3 is not doing `-nostdlib` either.

cmake/sdlcompilers.cmake
sezero: > That is indeed clang-cl.

I mean: arm32 linkage is failing most probably because you set the condition for MSVC_CLANG, but the compiler is simply MSVC without clang.

CMakeLists.txt
sezero: Missing libgcc is most possibly because clang has its own as compiler_rt (libclang_rt ??)

madebr: `-nostdlib` turned out to be needed for avoiding duplicate `_DllMainCRTStartup` symbol errors.

CMakeLists.txt
madebr: I ported this condition from SDL3's CMakeLists.txt:
```cmake
  # Note: The clang toolset for Visual Studio does not support /NODEFAULTLIB.
  if(MSVC AND NOT SDL_LIBC AND NOT MSVC_CLANG AND NOT SDL_CPU_ARM32)
    # Don't try to link with the default set of libraries.
    if(NOT WINDOWS_STORE)
      # FIXME: is this needed? "-nodefaultlib:MSVCRT" ia already added when SDL_LIBC is false
      target_link_options(SDL3-shared PRIVATE "/NODEFAULTLIB")
    endif()
  endif()
```

CMakeLists.txt
```diff
@@ -189,6 +194,18 @@ target_include_directories(SDL2
...
+  # The same applies to various functions normally belonging in the C library (for x86 architecture).
+  set_property(SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/SDL_mslibc.c" APPEND_STRING PROPERTY COMPILE_FLAGS " /GL-")
+endif()
+if(WIN32)
+  if(MSVC)
+    set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " /ENTRY:sdl2_compat_entry")
+  else()
+    set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -Wl,--entry=sdl2_compat_entry")
+  endif()
+endif()
```
madebr, line 208: I removed `-nostdlib` again, in favor for explicitly setting the entry function.
This fixes the following link error with msys2's clang32:
```
ld.lld: error: undefined symbol: ___udivdi3
>>> referenced by D:/a/sdl2-compat/sdl2-compat/src/sdl2_compat.c:1338
>>>               CMakeFiles/SDL2.dir/src/sdl2_compat.c.obj:(_Event3to2)
>>> referenced by D:/a/sdl2-compat/sdl2-compat/src/sdl2_compat.c:1381
>>>               CMakeFiles/SDL2.dir/src/sdl2_compat.c.obj:(_Event3to2)
>>> referenced by D:/a/sdl2-compat/sdl2-compat/src/sdl2_compat.c:1384
>>>               CMakeFiles/SDL2.dir/src/sdl2_compat.c.obj:(_Event3to2)

>>> referenced 2 more times
```

sezero: _DllCRTMain clash is not the reason for -nostdlib: We just do not want any libc / crt calls in the dll. ___udivdi3 etc should  be resolved by linking to clang's rtl (libclang_rt or whatever) just like with gcc linking to libgcc.

sezero: More importantly the generated dll is broken: note the linker warning:
```
D:/a/_temp/msys64/mingw32/bin/../lib/gcc/i686-w64-mingw32/13.1.0/../../../../i686-w64-mingw32/bin/ld.exe: warning: cannot find entry symbol sdl2_compat_entry; not setting start address
```
Because for x86 (and _possibly_ for other 32 bit windows such as arm32), the symbol is prefixed with an underscore.

I'd say not do this at all.

madebr: How do you explicitly ink to clang_rt?
The library has an arch dependent suffix: https://packages.msys2.org/package/mingw-w64-clang-i686-compiler-rt

sezero: Don't know. And, honestly didn't know that it'd be that painful.

sezero: First hits in [Google search](https://www.google.com/search?q=explicitly+link+to+clang_rt) shows that it _is pain_:
https://discourse.llvm.org/t/improve-autolinking-of-compiler-rt-and-libc-on-windows-with-lld-link/71392

CMakeLists.txt
```diff
@@ -273,11 +273,12 @@ else()
 endif()
 
 if(MINGW)
-  set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -nostdlib") # libgcc is needed for 32 bit (x86) builds
-  if(USE_GCC)
-    target_link_libraries(SDL2 PRIVATE "gcc")
-    set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -static-libgcc") # libgcc is needed for 32 bit (x86) builds
+  set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -nostdlib")
+  if(USE_CLANG)
+    target_compile_options(SDL2 PRIVATE "-rtlib=libgcc")
```
sezero, line 278: maybe add this only to LDFLAGS to avoid -Werror?

madebr: I'm not sure. This is a compiler flag for linux clang.
Don't you need this flag during compilation such that the compiler knows what runtime library to use?

slouken:
I haven't read the whole thread, but sdl2-compat should build with the same flags as SDL2, with no dependency on any C runtime on Windows. If there are targets (clang?) that require it, they should be considered unsupported and not build in CI.

CMakeLists.txt
```diff
@@ -275,10 +275,17 @@ endif()
...
+    if(SDL_CPU_X86)
+      target_link_libraries(SDL2 PRIVATE clang_rt.builtins-i386)
+    elseif(SDL_CPU_x64)
+      target_link_libraries(SDL2 PRIVATE clang_rt.builtins-x86_64)
+    else()
+      message(WARNING "Unknown clang_rt library (arch=${SDL_CPU_NAMES})")
+    endif()
+  else()
+    target_link_libraries(SDL2 PRIVATE "gcc")
+    set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -static-libgcc") # libgcc is needed for 32 bit (x86) builds
```
slouken, line 287: No, we should not link in static libgcc, we do not include the LGPL nor do we require that applications that use our code comply with the LGPL.

madebr: SDL3 does this. So it should be removed there too.

slouken: Yes, definitely.

slouken: Does using clang introduce any runtime dependencies or licensing requirements?

slouken:
FYI, I fixed the implicit memcpy() in 1dcd1cf.
We probably shouldn't bring over our own memcpy and memset, we should either call SDL3's versions or fix any implicit calls generated by the compiler.

CMakeLists.txt
sezero: Yes

slouken: What are they?

sezero: > Yes, definitely.

Then how will you resolve compiler runtime symbols?

> not include the LGPL nor do we require that applications that use our code comply with the LGPL.

IIRC, libgcc is LGPL3 but with "additional permissions described in the GCC Runtime Library Exception"
https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=COPYING.RUNTIME;hb=HEAD

sezero: > SDL3 does this. So it should be removed there too.

SDL2 does that too, _for years_

slouken: Ahhh, the runtime library exception is what we need here. What does it cover?

sezero: > Ahhh, the runtime library exception is what we need here. What does it cover?

Not a lawyer.. You should read that doc yourself.

madebr: [compiler-rt is Apache 2.0 with LLVM Exception] (https://packages.msys2.org/package/mingw-w64-clang-i686-compiler-rt?repo=clang32).

madebr:
> FYI, I fixed the implicit memcpy() in [1dcd1cf](https://github.com/libsdl-org/sdl2-compat/commit/1dcd1cf8850ce6ca359303fb35204c09c2e6e688). We probably shouldn't bring over our own memcpy and memset, we should either call SDL3's versions or fix any implicit calls generated by the compiler.

Not sure how to eliminate the implicit memcpy here:
https://github.com/libsdl-org/sdl2-compat/blob/1dcd1cf8850ce6ca359303fb35204c09c2e6e688/src/sdl2_compat.c#L5742-L5753

CMakeLists.txt
slouken: Okay, reading through it looks like both libgcc and compiler-rt are covered by exceptions that allow us to link them. Thank you for the clarification, we should be good to use them statically here.

slouken:
> Not sure how to eliminate the implicit memcpy here:
> 
> https://github.com/libsdl-org/sdl2-compat/blob/1dcd1cf8850ce6ca359303fb35204c09c2e6e688/src/sdl2_compat.c#L5742-L5753

Yeah, I don't know that we can. Returning a GUID on the stack probably isn't a great idea for other languages as well.
Maybe we should change the function to fill in a GUID parameter?

madebr:
> Yeah, I don't know that we can. Returning a GUID on the stack probably isn't a great idea for other languages as well. Maybe we should change the function to fill in a GUID parameter?

For SDL3, this can work.
But for sdl2-compat, you will still need to copy the guid to the stack memory of the caller when returning.

madebr:
Looks like `-lgcc -static-libgcc` is not sufficient for providing memcpy for arm32: https://github.com/libsdl-org/sdl2-compat/actions/runs/5779838735/job/15662676006?pr=93
So sdl2-compat now does both: `-lgcc -static-libgcc` + SDL3-based memcpy + memset implementation.

cmake/sdlcompilers.cmake
madebr: It was failing because the architecture was not detected. Now it does.

sezero:
> Looks like `-lgcc -static-libgcc` is not sufficient for providing memcpy for arm32: https://github.com/libsdl-org/sdl2-compat/actions/runs/5779838735/job/15662676006?pr=93

Well, obviously it cannot, at least because it's not gcc :)

>   SDL3-based memcpy + memset implementation.

yes, with a `#pragma intrinsic` (or `#pragma function`?) in mslibc.h, it should work

madebr:
> > Looks like `-lgcc -static-libgcc` is not sufficient for providing memcpy for arm32: https://github.com/libsdl-org/sdl2-compat/actions/runs/5779838735/job/15662676006?pr=93
> 
> Well, obviously it cannot, at least because it's not gcc :)

Dammit. You're right.
So `-gcc -static-libgcc` is not needed. But let's keep it for exotic untested mingw toolchains?

> > SDL3-based memcpy + memset implementation.
> 
> yes, with a `#pragma intrinsic` (or `#pragma function`?) in mslibc.h, it should work


I added it in `sdl2_compat.c` below the inclusion of `x86_msvc.h`.

sezero:
> > > Looks like `-lgcc -static-libgcc` is not sufficient for providing memcpy for arm32: https://github.com/libsdl-org/sdl2-compat/actions/runs/5779838735/job/15662676006?pr=93
> > 
> > 
> > Well, obviously it cannot, at least because it's not gcc :)
> 
> Dammit. You're right. So `-gcc -static-libgcc` is not needed. But let's keep it for exotic untested mingw toolchains?

Well, it is needed for at least x86 mingw

slouken:
Ah, for 64-bit division, that makes sense. Okay, I guess we keep it and use it whenever we're doing -nostdlib? We should document why it's necessary to save archeology work in the future.

sezero:
> > Not sure how to eliminate the implicit memcpy here:
> > https://github.com/libsdl-org/sdl2-compat/blob/1dcd1cf8850ce6ca359303fb35204c09c2e6e688/src/sdl2_compat.c#L5742-L5753
> 
> Yeah, I don't know that we can.

We can do the following:

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 5d78b5a..feb4b27 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -5747,7 +5747,8 @@ SDL_JoystickGetDeviceGUID(int idx)
     if (!jid) {
         SDL3_zero(guid);
     } else {
-        guid = SDL3_GetJoystickInstanceGUID(jid);
+        const SDL_JoystickGUID guid_ = SDL3_GetJoystickInstanceGUID(jid);
+        SDL3_memcpy(&guid, &guid_, sizeof(guid));
     }
     return guid;
 }
```

sezero:
Right now, this PR lokks pretty much OK. A few things to note:

- Do we need to apply https://github.com/libsdl-org/sdl2-compat/pull/93#issuecomment-1668096720 or something similar?

- The only windows target where we do not build with `/NODEFAULTLIB` is ARM 32 bits using MSVC, because we don't have any asm for it. (Any mingw or clang should be OK with `-nostdlib` because we link to libgcc.) Should we even care?

- I still think we should change the `MSVC_CLANG` check to simply `MSVC` for ARM 32 bit, i.e. https://github.com/libsdl-org/sdl2-compat/pull/93#pullrequestreview-1564191674 - but I won't fight for it.

- I think the -Wundef and -Wdocumentation fixes (except for the one-liner in sdl2_compat.c) are overkill here:  All of the affected sources/headers are directly copied from SDL2 where these fixes aren't applied at all. They should be fixed in SDL2 first if we do care. (Exception is the one-liner in sdl2_compat.c which is OK.)

- `SDL2COMPAT_WERROR` should default to OFF I guess - can be enabled in CI..

CMakeLists.txt
```diff
@@ -263,16 +268,28 @@ else()
...
 if(MINGW)
-  target_link_libraries(SDL2 PRIVATE -nostdlib)
-  target_link_libraries(SDL2 PRIVATE -static-libgcc -lgcc) # libgcc is needed for 32 bit (x86) builds
+  set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -nostdlib")
+  if(USE_CLANG)
+    if(SDL_CPU_X86)
+      target_link_libraries(SDL2 PRIVATE clang_rt.builtins-i386)
+    endif()
+  else()
+    if(SDL_CPU_X86 OR SD_CPU_ARM32)
```
sezero, line 277: Typo yes? `SDL_CPU_ARM32`

madebr: Thanks :)

sezero:
> > ```
> > * The only windows target where we do not build with `/NODEFAULTLIB` is ARM 32 bits using MSVC, because we don't have any asm for it. (Any mingw or clang should be OK with `-nostdlib` because we link to libgcc.) Should we even care?
> > ```
> 
> This question applies to SDL3 as well.

And SDL2 too. @slouken: Do we keep going as is?

> > ```
> > * I think the -Wundef and -Wdocumentation fixes are overkill here:  All of the affected sources/headers are directly copied from SDL2 where these fixes aren't applied at all. They should be fixed in SDL2 first if we do care. (Exception is the one-liner in sdl2_compat.c which is OK.)\
> > ```
> 
> The `-Wundef` warnings are already fixed in SDL3, and are in pr for SDL2 at [libsdl-org/SDL#7554](https://github.com/libsdl-org/SDL/pull/7554). About the `Wdocumentation` patches, let's port these to SDL2.

OK, as you wish.

sezero:
Off-topic actually, but since you are went overboard with this, how about
testing SDL3 symbol correctness automatically? I mean this:
https://github.com/libsdl-org/sdl2-compat/blob/main/src/sdl2_compat.c#L123-L132

I have been doing it manually by changing that `#if 0` to `#if 1` successfully,
and if the compiler is a c++ compiler, it fails immediately (in plain C, it'll
be just a warning): Try changing the `Init` parameter from Uint32 to Sint32 in
sdl3_syms.h and see. Useful for detecting SDL3 api changes not being properly
reflected here.

For automatic testing, I can think of something like this: Do you think that we
should care about that and add something equivalent to the cmake'ry?

```diff
diff --git a/src/sdl3_testsyms.cc b/src/sdl3_testsyms.cc
new file mode 100644
index 0000000..a389bf2
--- /dev/null
+++ b/src/sdl3_testsyms.cc
@@ -0,0 +1 @@
+#include "sdl2_compat.c"
diff --git a/src/Makefile.linux b/src/Makefile.linux
index 86ce95b..eeaf23c 100644
--- a/src/Makefile.linux
+++ b/src/Makefile.linux
@@ -6,2 +6,3 @@ INCLUDES = -Iinclude
 CC = gcc
+CXX= g++
 LD = $(CC)
@@ -25,3 +26,3 @@ all: $(SHLIB)
 
-$(SHLIB): $(OBJ)
+$(SHLIB): $(OBJ) sdl3_testsyms.o
 	$(LD) -o $@ $(LDFLAGS) $(OBJ) $(LDLIBS)
@@ -34,2 +35,5 @@ $(SHLIB): $(OBJ)
 
+sdl3_testsyms.o: sdl3_testsyms.cc
+	$(CXX) $(CFLAGS) $(CPPFLAGS) -DSDL2COMPAT_TEST_SYMS=1 $(INCLUDES) -o $@ -c $<
+
 distclean: clean
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index bb0b33a..1b8af9c 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -122,3 +122,6 @@ extern "C" {
  */
-#if 0
+#ifndef SDL2COMPAT_TEST_SYMS
+#define SDL2COMPAT_TEST_SYMS 0
+#endif
+#if SDL2COMPAT_TEST_SYMS
 #define SDL3_SYM(rc,fn,params,args,ret) \
```

sezero:
A little late, but noticed that we are now building with `-fno-strict-aliasing`: It doesn't seem necessary here in sdl2-compat. (It was only added to SDL as a temporary measure for https://github.com/libsdl-org/SDL/issues/2974)

slouken:
> > ```
> > * The only windows target where we do not build with `/NODEFAULTLIB` is ARM 32 bits using MSVC, because we don't have any asm for it. (Any mingw or clang should be OK with `-nostdlib` because we link to libgcc.) Should we even care?
> > ```
> 
> This question applies to SDL3 as well.

Should we even care that we're not using /NODEFAULTLIB? That's actually the reason there isn't ARM support in the Visual Studio project files.

I don't know of anyone shipping an ARM version of SDL though, so I think it's fine for us to verify in CI and then let people build it themselves with their own C runtime dependencies if they want.

sezero:
@madebr: Regarding the `Add SDL3-based implementation of memcpy and memcpy` commit:
I believe the following is the actual intention which is what SDL3 and SDL2 do.
OK to push, or am I missing anything?

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index df5302e..72e9c3b 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -589,44 +589,54 @@ LoadSDL3(void)
             }
         }
     }
     return okay;
 }
 
-#ifndef SDL_BUILDING_WINRT
-#if defined(_MSC_VER) && defined(_M_IX86)
+#if defined(_MSC_VER) && !defined(SDL_BUILDING_WINRT)
+#ifdef _M_IX86
 #include "x86_msvc.h"
 #endif
 
-#if defined(_WIN32) && !defined(__WATCOMC__)
 /* NOLINTNEXTLINE(readability-redundant-declaration) */
 extern void *memcpy(void *dst, const void *src, size_t len);
 /* NOLINTNEXTLINE(readability-redundant-declaration) */
 extern void *memset(void *dst, int c, size_t len);
-#ifdef _MSC_VER
 #ifndef __INTEL_LLVM_COMPILER
 #pragma intrinsic(memcpy)
 #pragma intrinsic(memset)
 #endif
 #pragma function(memcpy)
 #pragma function(memset)
-#endif
 
 /* NOLINTNEXTLINE(readability-inconsistent-declaration-parameter-name) */
 void *memcpy(void *dst, const void *src, size_t len)
 {
     return SDL3_memcpy(dst, src, len);
 }
 
 /* NOLINTNEXTLINE(readability-inconsistent-declaration-parameter-name) */
 void *memset(void *dst, int c, size_t len)
 {
     return SDL3_memset(dst, c, len);
 }
+#endif  /* MSVC && !WINRT */
+
+#if defined(__ICL) && defined(_WIN32)
+/* The classic Intel compiler generates calls to _intel_fast_memcpy
+ * and _intel_fast_memset when building an optimized SDL library */
+void *_intel_fast_memcpy(void *dst, const void *src, size_t len)
+{
+    return SDL_memcpy(dst, src, len);
+}
+
+void *_intel_fast_memset(void *dst, int c, size_t len)
+{
+    return SDL_memset(dst, c, len);
+}
 #endif
-#endif  /* ! WINRT */
 
 #ifdef SDL_BUILDING_WINRT
 EXTERN_C void error_dialog(const char *errorMsg);
 #elif defined(_WIN32)
 static void error_dialog(const char *errorMsg)
 {
```

madebr:
That looks sensible. Untested, but ok by me.

madebr:
I think you want to use `SDL3_memcpy` instead of `SDL_memcpy` in the new mem* functions.

PR #96, opened 2023-08-08
Verify sdl3 symbols by building sdl2_compat in c++ mode and defining SDL2COMPAT_TEST_SYMS

madebr:
Also fix a `-Wsign-compare` warning that shows up in c++ mode.


sezero (approved):
This looks good to me.

P.S.: We should apply the -Wsign-compare fix to SDL2 too.

madebr:
I also added schedule to the workflow event trigger list.
https://github.com/libsdl-org/sdl2-compat/pull/81#issuecomment-1652225095

sezero:
OK, let's get this in.

PR #105, opened 2023-11-03
simplifications after SDL3 bool type change.

sezero:
@slouken, @icculus: Please review carefully so that I didn't make any logic errors.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -399,7 +399,7 @@ SDL2Compat_GetHintBoolean(const char *name, SDL_bool default_value)
         return default_value;
     }
 
-    return (SDL3_atoi(val) != 0) ? SDL_TRUE : SDL_FALSE;
+    return (SDL3_atoi(val) != 0);
 }
 
 
@@ -2188,7 +2188,7 @@ SDL_LoadFile_RW(SDL2_RWops *rwops2, size_t *datasize, int freesrc)
     void *retval = NULL;
     SDL_RWops *rwops3 = RWops2to3(rwops2);
     if (rwops3) {
-        retval = SDL3_LoadFile_RW(rwops3, datasize, freesrc ? SDL_TRUE : SDL_FALSE);
+        retval = SDL3_LoadFile_RW(rwops3, datasize, freesrc != 0);
         if (!freesrc) {
             SDL3_DestroyRW(rwops3);  /* don't close it because that'll close the SDL2_RWops. */
         }
@@ -2211,7 +2211,7 @@ SDL_LoadWAV_RW(SDL2_RWops *rwops2, int freesrc, SDL2_AudioSpec *spec2, Uint8 **a
         SDL_RWops *rwops3 = RWops2to3(rwops2);
         if (rwops3) {
             SDL_AudioSpec spec3;
-            const int rc = SDL3_LoadWAV_RW(rwops3, freesrc ? SDL_TRUE : SDL_FALSE, &spec3, audio_buf, audio_len);
+            const int rc = SDL3_LoadWAV_RW(rwops3, freesrc != 0, &spec3, audio_buf, audio_len);
             SDL3_zerop(spec2);
             if (rc == 0) {
                 spec2->format = spec3.format;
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -2211,7 +2211,7 @@ SDL_LoadWAV_RW(SDL2_RWops *rwops2, int freesrc, SDL2_AudioSpec *spec2, Uint8 **a
         SDL_RWops *rwops3 = RWops2to3(rwops2);
         if (rwops3) {
             SDL_AudioSpec spec3;
-            const int rc = SDL3_LoadWAV_RW(rwops3, freesrc ? SDL_TRUE : SDL_FALSE, &spec3, audio_buf, audio_len);
+            const int rc = SDL3_LoadWAV_RW(rwops3, freesrc, &spec3, audio_buf, audio_len);
```
slouken, line 2214: This should probably be `(freesrc != 0)`

src/sdl2_compat.c
```diff
@@ -3377,10 +3377,7 @@ SDL_RenderTargetSupported(SDL_Renderer *renderer)
     if (ret == 0) {
         /* SDL_RENDERER_TARGETTEXTURE was removed in SDL3, check by name for 
          * renderer that does not support render target. */
-        if (SDL3_strcmp(info.name, "opengles") == 0) {
-            return SDL_FALSE;
-        }
-        return SDL_TRUE;
+        return !!SDL3_strcmp(info.name, "opengles");
```
slouken, line 3380: This probably shouldn't be changed, for clarity.

src/sdl2_compat.c
```diff
@@ -5257,7 +5254,7 @@ SDL_Has3DNow(void)
         cpuid(0x80000000, a, b, c, d);
         if ((unsigned int)a >= 0x80000001) {
             cpuid(0x80000001, a, b, c, d);
-            return (d & 0x80000000) ? SDL_TRUE : SDL_FALSE;
+            return !!(d & 0x80000000);
```
slouken, line 5257: I prefer the explicit ternary instead of the `!!` syntax. It's easy to think it's a logical not if you're scanning quickly.

src/sdl2_compat.c
```diff
@@ -5374,7 +5371,7 @@ DECLSPEC SDL_Window * SDLCALL
 SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
 {
     SDL_Window *window = NULL;
-    SDL_bool hidden = (flags & SDL_WINDOW_HIDDEN) ? SDL_TRUE : SDL_FALSE;
+    const SDL_bool hidden = flags & SDL_WINDOW_HIDDEN;
```
slouken, line 5374: `(flags & SDL_WINDOW_HIDDEN) != 0`
Remember, this is actually supposed to be a 0 or 1 value, not 0 or non-zero.

src/sdl2_compat.c
```diff
@@ -5436,13 +5433,7 @@ SDL_CreateShapedWindow(const char *title, unsigned int x, unsigned int y, unsign
...
 SDL_SetWindowFullscreen(SDL_Window *window, Uint32 flags)
 {
-    SDL_bool flags3 = SDL_FALSE;
-
-    if (flags & SDL2_WINDOW_FULLSCREEN_DESKTOP) {
-        flags3 = SDL_TRUE;
-    }
-
-    return SDL3_SetWindowFullscreen(window, flags3);
+    return SDL3_SetWindowFullscreen(window, !!(flags & SDL2_WINDOW_FULLSCREEN_DESKTOP));
```
slouken, line 5436: `(flags & SDL2_WINDOW_FULLSCREEN_DESKTOP) != 0`

src/sdl2_compat.c
sezero: Changed it and all its similars

src/sdl2_compat.c
sezero: OK, dropped

src/sdl2_compat.c
sezero: It's only locally used, but changed as suggested

src/sdl2_compat.c
sezero: Changed as suggested

sezero:
I think I covered all change requests: Please re-review.

src/sdl2_compat.c
sezero: Done

slouken (approved):

slouken:
Looks good!

Issue #106, opened 2023-11-05
textinput, textedit and drop events broken after recent SDL3 changes


sezero:
Just run checkkeys test program and hit a few keys.

CC: @slouken, @icculus

sezero:
For `Event3to2` I can do the following and it does fix checkkeys. It also fixes
e.g. quakespasm text input in game console.
```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 3749523..ffec361 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1388,6 +1388,14 @@ Event3to2(const SDL_Event *event3, SDL2_Event *event2)
     SDL3_memcpy((&event2->common) + 1, (&event3->common) + 1, sizeof (SDL2_Event) - sizeof (SDL2_CommonEvent));
     /* mouse coords became floats in SDL3: */
     switch (event3->type) {
+    case SDL_EVENT_TEXT_INPUT:
+        SDL3_strlcpy(event2->text.text, event3->text.text, sizeof(event2->text.text));
+        break;
+    case SDL_EVENT_TEXT_EDITING:
+        SDL3_strlcpy(event2->edit.text, event3->edit.text, sizeof(event2->edit.text));
+        event2->edit.start = event3->edit.start;
+        event2->edit.length = event3->edit.length;
+        break;
     case SDL_EVENT_MOUSE_MOTION:
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->motion.windowID));
         if (renderer) {
```

However: What to do for `Event2to3` ?
Should we `SDL3_malloc` the text and feed to SDL3? If so, where will it be freed
and whose job will it be to free it?

P.S.: There may be more issues with event changes other than those two I mentioned.

sezero:
> P.S.: There may be more issues with event changes other than those two I mentioned.

DropEvent should be broken by the looks of it. Won't be testing or working on it myself

slouken:
> However: What to do for `Event2to3` ? Should we `SDL3_malloc` the text and feed to SDL3? If so, where will it be freed and whose job will it be to free it?

We should use SDL_AllocateEventMemory(), and it'll be automatically freed later.

sezero:
OK, will be trying

sezero:
Current progress below. Will need to do something for DropEvent too, but I'll have to read more to handle that.

```diff
[...]

PR #107, opened 2023-11-06
fix text input, text edit and drop events after recent SDL3 changes

sezero:
Fixes: https://github.com/libsdl-org/sdl2-compat/issues/106.

Changes are in `Event3to2` and `Event2to3` procedures. `Event2to3` handles
user program to SDL event pushes: SDL_TEXTEDITING_EXT isn't handled there
as discussed in #106, and I assumed the same for drop events shouldn't be
handled there either.

@slouken: Please review carefully.


slouken:
I wouldn't expect the application to be pushing `SDL_EVENT_TEXT_EDITING` either, but otherwise it looks great.

slouken (approved):

sezero:
OK, commented out that part and pushed.

PR #108, opened 2023-11-09
Update for SysWM properties

Kontrabant:


sezero:
Here is a patch (on top of yours) to update sdl3_include_wrapper.h: [patch.txt](https://github.com/libsdl-org/sdl2-compat/files/13305317/patch.txt)

src/sdl2_compat.h
```diff
@@ -194,7 +339,7 @@ typedef enum
   SDL2_SYSWM_RISCOS
 } SDL2_SYSWM_TYPE;
 
-struct SDL2_SysWMinfo
+struct SDL_SysWMinfo
```
madebr, line 342: Can we avoid the typedefs and includes by making `SDL_SysWMinfo` use void pointers instead of typedefs? Or are these important?
It looks like `SDL_GetWindowWMInfo` is simply forwarding pointers and ints.

Kontrabant: I think we can. It will be a bit ugly, but the SDL2 syswm struct should be forever frozen in time and hopefully this code will never have to be touched again.

madebr: And you already did the hard work of adding the types. So it would be a waste of reverting this.

sezero: >  It will be a bit ugly,

It did get ugly:
```diff
+      wminfo->info.x11.window = (unsigned long)(uintptr_t)SDL3_GetProperty(props, "SDL.window.x11.window");
```

Kontrabant: Losing 30 seconds of copy-paste from the original headers isn't a big deal :)

I like this much better without all of the `#ifdef` mess anyways.  We just need to keep the bit for the Cocoa and UIKit auto refcounting.

Kontrabant:
> Here is a patch (on top of yours) to update sdl3_include_wrapper.h: [patch.txt](https://github.com/libsdl-org/sdl2-compat/files/13305317/patch.txt)

Applied, thanks!

src/sdl2_compat.c
```diff
@@ -281,7 +281,7 @@ static char loaderror[256];
     }
 #elif defined(__unix__)
     #include <dlfcn.h>
-    #define SDL3_LIBNAME "libSDL3.so.0"
+    #define SDL3_LIBNAME "/home/franz/Documents/sdl-git/SDL/cmake-build-debug/libSDL3.so.0"
```
slouken, line 284: You probably don't want to commit this. :)

src/sdl2_compat.c
```diff
@@ -2491,63 +2491,46 @@ SDL_GetTicks64(void)
...
+    }
+
+    if (!driver) {
+        SDL_SetError("No video driver initialized");
+        return SDL_FALSE;
+    }
+
+    if (SDL_strcmp(driver, "Android") == 0) {
+        wminfo->subsystem = SDL2_SYSWM_ANDROID;
+        wminfo->info.android.window = (void *)SDL3_GetProperty(props, "SDL.window.android.window");
```
slouken, line 2510: The cast to (void *) isn't needed, here or below.

src/sdl2_compat.c
madebr: I'm looking into adding rpaths during build time (not install time) such that changes like these are not required anymore.

Kontrabant: Whoops, no I did not :)

src/sdl2_compat.c
Kontrabant: Fixed

sezero:
What else is left?

slouken (approved):

sezero:
.. and I guess we can't do anything about SDL_SYSWMEVENT events? remove those commented out stuff?

Kontrabant:
The last batch of changes introduced `SDL.internal` namespaces for internal properties, sdl2-compat should use these for tracking popup parents. Might as well fix that up now.

slouken:
> .. and I guess we can't do anything about SDL_SYSWMEVENT events? remove those commented out stuff?

Actually, we can. We can use SDL_SetWindowsMessageHook() and SDL_SetX11EventsHook() to set callbacks to generate SDL_SYSWMEVENT events when they're enabled.

Kontrabant:
> > .. and I guess we can't do anything about SDL_SYSWMEVENT events? remove those commented out stuff?
> 
> Actually, we can. We can use SDL_SetWindowsMessageHook() and SDL_SetX11EventsHook() to set callbacks to generate SDL_SYSWMEVENT events when they're enabled.

Let's tackle that separately since this is otherwise good to go.

PR #109, opened 2023-11-11
Updated for SDL3 changes to gamepad face buttons

slouken:
Updated for SDL3 changes to gamepad face buttons

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1162,6 +1162,8 @@ static SDL_mutex *EventWatchListMutex = NULL;
 static EventFilterWrapperData *EventWatchers2 = NULL;
 static SDL_JoystickID *joystick_list = NULL;
 static int num_joysticks = 0;
+static SDL_JoystickID *gamepad_button_swap_list = NULL;
+static int num_gamepad_button_swap_list = 0;
 static SDL_SensorID *sensor_list = NULL;
 static int num_sensors = 0;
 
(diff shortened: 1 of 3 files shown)


src/sdl2_compat.c
```diff
@@ -1350,6 +1352,79 @@ SDL_LOG_IMPL(Error, ERROR)
 SDL_LOG_IMPL(Critical, CRITICAL)
 #undef SDL_LOG_IMPL
 
+/* !!! FIXME: when we override SDL_Quit(), we need to free/reset gamepad_button_swap_list */
```
sezero, line 1355: We override SDL_Quit already (for gesture stuff)

slouken: Great, I added cleanup for that.

sezero: Should we do that cleanup after SDL3_Quit like you do, or before?

src/sdl2_compat.c
```diff
@@ -3544,6 +3623,24 @@ SDL_Quit(void)
...
     SDL3_Quit();
+
+    if (joystick_list) {
+        SDL_free(joystick_list);
+        joystick_list = NULL;
+    }
+    num_joysticks = 0;
+
+    if (sensor_list) {
+        SDL_free(sensor_list);
```
sezero, line 3634: `SDL3_free` instead of `SDL_free` (same for the ones below)

slouken: Fixed

src/sdl2_compat.c
slouken: My intuition is that there may be events in SDL_Quit() and we might need the data there, but I'll move it beforehand for consistency with everything else for now.

sezero: OK then

sezero (approved):

PR #111, opened 2023-11-13
SDL_CreateWindowFrom() now takes a set of properties that describe the native window and options.

slouken:
sdl2-compat changes for https://github.com/libsdl-org/SDL/pull/8548


sezero (approved):

sezero:
FYI: Pushed https://github.com/libsdl-org/sdl2-compat/commit/b43c7a9985d87d0bc16202a06082f6b265760ddf to silence a strict-aliasing warning.

PR #113, opened 2023-11-21
add initial Android.mk and a build script for ndk-build.

sezero:


build-scripts/androidbuild.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+# Usage: androidbuildlibs.sh [arg for ndk-build ...]"
```
slouken, line 2: androidbuild.sh

slouken:
This seems reasonable, but there's a bunch of Java code involved in integration with SDL. You'd need to include SDL3 Java code, which would look for SDL3 symbols. Have you tried this? Does it work?

build-scripts/androidbuild.sh
sezero: fixed

sezero:
> This seems reasonable, but there's a bunch of Java code involved in integration with SDL. You'd need to include SDL3 Java code, which would look for SDL3 symbols. Have you tried this? Does it work?

No, build-tested only: it succeeds generating a libSDL2.so, no errors or warnings, and that's it. (I don't have anything to test.)

madebr:
As Sam wrote, the java code is tightly connected with the so.
It's even [version dependent](https://github.com/libsdl-org/SDL/blob/f33ab44f8469a427e7fd7c7552b91776503ec1ce/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L349-L356).

I think there are 2 use cases:
- SDL2 source games. These projects need sdl2-compat compatible java sources, which can probably be copied from SDL2. sdl2-compat will also need to provide a SDL2-compatible `JNI_GetEnv`. I think we need to copy the SDL2 java source codes to sdl2-compat.
- SDL2 binary games. This would require all of the above. Replacing SDL2.so is not sufficient, as the java sources are version dependent. So these would require the dex files to be decompiled to java, the SDL2 java sources replaced with those form sdl2-compat and then recompiled again to dex. Replacing the java sources can be easy if the SDL2 java library is dex'ed seperately as e.g. `classes[0-9]*.dex`.

Since SDL2 also does not support SDL updates by replacing SDLx.so, I think we should not care about the last case.

sezero:
OK.

I will leave this open for now, so feel free to add any changes to my branch.

slouken:
I don't think this is a use case we can reasonably support.

sezero:
> I don't think this is a use case we can reasonably support.

Shall I close then?

Although: Why do we have the Android.mk and associated androidbuild[libs].sh files in SDL2 and SDL3 (and even test them in CI)?

madebr:
> > I don't think this is a use case we can reasonably support.
> 
> Shall I close then?

I'm not sure whether he was refering to my last use case.
The first one looks reasonable.

Asking @1bsyl  for feedback.

icculus:
> SDL2 binary games

There's no universe where we are replacing SDL in these in the field on Android, even without the Java concerns.

The use-case on Android is someone trying to get their own game updated, when an unmaintained SDL2 has become unpleasant/unusable on a future Android release, without updating from the SDL2 API.

For this, we can do whatever we want, including using only the SDL3 Java, and chaining an SDL_main in sdl2-compat to the app's real SDL_main...or something along those lines.

1bsyl:
I don´t think sdl2-compat should be used for android, 
it's not a platform where we re-package .apk ..

of course, people can still take an old .so, for sdl2, and use we sdl3+sdl2-compat, and old asset.. create a new apk.. this can probably work
but well this need lot of manip, an probably a really custom android makefile for that

slouken:
Let's go ahead and merge what you have, since it'll certainly be needed by anyone building for Android, and we can defer any other Java related work until we have a use case to solve.

sezero:
OK then

PR #121, opened 2023-12-20
Fix invalid read: displayID wasn't set

1bsyl:


sezero (approved):
This should be OK. @icculus and/or @slouken?

slouken (approved):

PR #128, opened 2024-01-10
build: Include vendor info in the return from SDL_GetRevision()

smcv:
This matches "classic" SDL2 behaviour. For example, when built from a distro package, we would like to see the packaging version number in the SDL_GetRevision() and in strings(1) output, which will let a developer track down whether any patches were included in the distro build that they're using.

Before this commit, SDL_revision.h did include the SDL2COMPAT_VENDOR_INFO in its version info, so testver would output something like:

    INFO: Compiled version: 2.28.50 (SDL-2.28.50-no-vcs (Debian 2.28.50~git20240110~c5faaac+ds-1))

but SDL_GetRevision() did not return the SDL2COMPAT_VENDOR_INFO, so testver would output something like:

    INFO: Linked version: 2.28.50 (SDL-2.28.50-no-vcs)

This commit changes that output to be the same as the "Compiled version", for more informative version information in the presumably common case where a game was compiled against classic SDL2 but then run with sdl2-compat.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,12 @@ else()
   set(SDL_REVISION "SDL-${SDL_VERSION}-no-vcs")
 endif()
 
+if(SDL2COMPAT_VENDOR_INFO)
+  set(SDL2COMPAT_REVISION "${SDL_REVISION} (${SDL2COMPAT_VENDOR_INFO})")
+else()
+  set(SDL2COMPAT_REVISION "${SDL_REVISION}")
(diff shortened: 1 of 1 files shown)


smcv:
> SDL-2.28.50-no-vcs

We see this because distros' build systems often do not do binary builds directly from git: instead, the package maintainer provides a tarball with source code and some build instructions (which they probably maintain using git, but that's an implementation detail), and then the automated build system builds from that tarball using the build instructions.

PR #132, opened 2024-01-23
test: Extend testautomation timeout to match classic SDL2

smcv:
With the 40-second limit it times out on Debian armel (ARMv5 softfloat) autobuilders, and on some older/unsupported "ports" architectures.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -155,7 +155,7 @@ if(HAVE_OPENGLES_V2)
   endif()
 endif()
 
-test_program(testautomation NONINTERACTIVE TIMEOUT 40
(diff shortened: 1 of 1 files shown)


smcv:
I don't know for sure that this will pass on all reasonable armel machines, but the version in SDL 2.29.1 took about 1 minute on an official Debian autobuilder, so 2 minutes will probably be enough.

Of course, I would encourage anyone who is interested in actually playing a game to get better hardware - Raspberry Pi version 2 or better can run Debian armhf, which is ARMv7 hardfloat, a much more reasonable thing to be playing games on. But continuing to compile SDL on armel gives us a convenient example of an architecture where atomic instructions aren't available, so it's probably a useful stand-in for some of the embedded/console architectures that SDL can target.

PR #138, opened 2024-02-11
Updated for structure renaming in SDL3

slouken:


sezero:
Is SDL_Event still compatible?

EDIT: I tested quakespasm and seems to work, so far...

slouken:
Yes, we copy the rest of the payload after the timestamp, so it should be fine.

PR #139, opened 2024-02-12
Update for property define renaming in SDL3

zturtleman:
Update for SDL3 changes in https://github.com/libsdl-org/SDL/pull/9049.


slouken:
Merged, thanks!

Issue #144, opened 2024-03-07
failures after SDL3 enum use changes (SDL_PixelFormatEnum, in particular)


sezero:
After enum patch merges of @Susko3 in SDL3, `SDL_PixelFormatEnum` in
particular, sdl2-compat needed adjusting. I did the following to fix
the SDL3 signatures:

```diff
diff --git a/src/sdl3_syms.h b/src/sdl3_syms.h
index 35234d9..a1c78f8 100644
--- a/src/sdl3_syms.h
+++ b/src/sdl3_syms.h
@@ -280,10 +280,10 @@ SDL3_SYM_RENAMED(int,CondSignal,SignalCondition,(SDL_Condition *a),(a),return)
 SDL3_SYM_RENAMED(int,CondBroadcast,BroadcastCondition,(SDL_Condition *a),(a),return)
 SDL3_SYM_RENAMED(int,CondWait,WaitCondition,(SDL_Condition *a, SDL_Mutex *b),(a,b),return)
 SDL3_SYM(int,WaitConditionTimeout,(SDL_Condition *a, SDL_Mutex *b, Sint32 c),(a,b,c),return)
-SDL3_SYM(const char*,GetPixelFormatName,(Uint32 a),(a),return)
-SDL3_SYM_RENAMED(SDL_bool,PixelFormatEnumToMasks,GetMasksForPixelFormatEnum,(Uint32 a, int *b, Uint32 *c, Uint32 *d, Uint32 *e, Uint32 *f),(a,b,c,d,e,f),return)
-SDL3_SYM_RENAMED(Uint32,MasksToPixelFormatEnum,GetPixelFormatEnumForMasks,(int a, Uint32 b, Uint32 c, Uint32 d, Uint32 e),(a,b,c,d,e),return)
-SDL3_SYM_RENAMED(SDL_PixelFormat*,AllocFormat,CreatePixelFormat,(Uint32 a),(a),return)
+SDL3_SYM(const char*,GetPixelFormatName,(SDL_PixelFormatEnum a),(a),return)
+SDL3_SYM_RENAMED(SDL_bool,PixelFormatEnumToMasks,GetMasksForPixelFormatEnum,(SDL_PixelFormatEnum a, int *b, Uint32 *c, Uint32 *d, Uint32 *e, Uint32 *f),(a,b,c,d,e,f),return)
[...]

PR #145, opened 2024-03-07
Updated for SDL3 enum changes

slouken:


slouken:
This should fix https://github.com/libsdl-org/sdl2-compat/issues/144

Issue #143, opened 2024-03-06
changes needed for SDL_WindowsMessageHook signature changes in SDL3


sezero:
At present, we don't have an SDL2_WindowsMessageHook and we pass-through the related calls which is wrong.

PR #146, opened 2024-03-13
Fix SDL_SetWindowsMessageHook() after SDL3 changes.

sezero:
Fixes https://github.com/libsdl-org/sdl2-compat/issues/143

@slouken: Please review. Returning SDL_TRUE is the correct behavior, yes?


slouken (approved):
Yes, this looks good.

PR #150, opened 2024-03-28
Updated for latest SDL3 changes as of d00ccc1546c0943481f5d65459496a4…

slouken:
…7fe9721df


sezero (approved):
Is `SDL3_TextInputActive` the right thing to do here?

Anyways, won't argue. OK by me.

slouken:
Actually I think SDL_FALSE is the right thing to do here. I'll update the PR and merge.

PR #153, opened 2024-03-31
Handle distinction between full-screen and full-screen desktop

cgutman:
The existing code would always enter full-screen exclusive mode if SDL_SetWindowDisplayMode() was called before SDL_SetWindowFullscreen(), even if the flag value was SDL_WINDOW_FULLSCREEN_DESKTOP.

Likewise, the code would always enter full-screen desktop mode if SDL_SetWindowDisplayMode() wasn't called, even if the caller passed SDL_WINDOW_FULLSCREEN to ask for full-screen exclusive mode.

Since SDL3 uses the presence of a full-screen mode to tell whether to use full-screen exclusive or full-screen desktop mode, we will stash the provided display mode in a property until we know whether the caller actually wants modesetting or not.

While I was here modifying this code, I also fixed some `SDL3_GetPrimaryDisplay()` calls that should have been using `SDL3_GetDisplayForWindow()` instead.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -5673,6 +5673,36 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
     return 0;
 }
 
+#define PROP_WINDOW_FULLSCREEN_MODE "sdl2-compat.window.fullscreen-mode"
+
+static int
+ApplyFullscreenMode(SDL_Window *window)
+{
+    SDL_DisplayMode *mode = (SDL_DisplayMode *) SDL3_GetProperty(SDL3_GetWindowProperties(window), PROP_WINDOW_FULLSCREEN_MODE, NULL);
+    if (mode) {
+        /* Always try to enter fullscreen on the current display */
+        mode->displayID = SDL3_GetDisplayForWindow(window);
+    }
+    if (mode && SDL3_SetWindowFullscreenMode(window, mode) == 0) {
+        return 0;
+    } else {
+        int count = 0;
+        const SDL_DisplayMode **list;
+        int ret;
+
+        /* FIXME: at least set a valid fullscreen mode */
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -5673,6 +5673,42 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
...
 
+#define PROP_WINDOW_FULLSCREEN_MODE "sdl2-compat.window.fullscreen-mode"
+
+static int
+ApplyFullscreenMode(SDL_Window *window)
+{
+    SDL_DisplayMode *mode = (SDL_DisplayMode *) SDL3_GetProperty(SDL3_GetWindowProperties(window), PROP_WINDOW_FULLSCREEN_MODE, NULL);
+    if (mode) {
+        /* Always try to enter fullscreen on the current display */
+        mode->displayID = SDL3_GetDisplayForWindow(window);
```
cgutman, line 5684: Not sure what we're supposed to do if the current display isn't compatible with whatever mode the application wanted (for example, they enumerated modes for a different monitor which isn't present on the current monitor). Since SDL2 doesn't have a `displayID` field in the mode, we're forced to guess which monitor they want to use.

We could do one (or a combination) of the following:
1. Use the display of the window at the time the modeset is happening
2. Use the display of the window at the time `SDL_SetWindowDisplayMode()` is called
3. Enumerate modes on all displays and try to find the monitor that the caller's mode came from (which is quite possibly not unique to a single monitor).

Right now I'm doing 1 because it seems like probably what developers will expect and avoids the possibilty of the window flying across to another monitor when it enters fullscreen mode (as can happen with 2 or 3).

slouken: This seems fine. For SDL3 the mode includes the display so the behavior is that the monitor is the display in the mode that was set, but for SDL2 the current window monitor is probably expected.

PR #156, opened 2024-04-02
Only scale mouse coordinates if logical size is set

cgutman:
SDL2 only scales mouse coordinates to the renderer's (logical) size if `SDL_RenderSetLogicalSize()` was called to enable logical scaling. If not, it will send unmodified coordinates in mouse events. Relevant SDL2 code: https://github.com/libsdl-org/SDL/blob/f1690e265e306818882c7c876a9e85492eeefa42/src/render/SDL_render.c#L761-L787

Unconditionally calling `SDL_ConvertEventToRenderCoordinates()` doesn't exactly match SDL2's behavior here because it will still adjust for DPI even if the renderer's logical presentation size is not set. I'm assuming `SDL_ConvertEventToRenderCoordinates()` is working as intended, so we have to check if logical scaling is enabled before using it.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1572,9 +1572,13 @@ Event3to2(const SDL_Event *event3, SDL2_Event *event2)
     case SDL_EVENT_MOUSE_MOTION:
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->motion.windowID));
         if (renderer) {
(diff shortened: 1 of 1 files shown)


cgutman:
CI needs to be rerun after cde306383b741e1d5a2b8b49e74ef0b316a5336f landed. I don't have permission to kick it myself (without pushing again).

slouken:
Merged, thanks!

PR #161, opened 2024-05-07
update after SDL3 error api changes

sezero:


src/sdl2_compat.c
```diff
@@ -2400,7 +2424,7 @@ stdio_seek(SDL2_RWops *rwops2, Sint64 offset, int whence)
         }
         return pos;
     }
-    return SDL3_Error(SDL_EFSEEK);
+    return SDL3_SetError("Error seeking in datastream");
```
slouken, line 2427: These could call the SDL2 version of SDL_Error() if you wanted.

slouken (approved):

PR #162, opened 2024-05-09
update after SDL3 changing SDL_WindowFlags to 64bit

sezero:
also fixes SDL3_CreateRenderer having the wrong type for its flags param

@slouken: I hope I didn't miss anything wrt 32 <-> 64 bit conversion of
flags between SDL2 and SDL3, but not 100% sure. Please review carefully.


src/sdl2_compat.c
```diff
@@ -6161,7 +6161,7 @@ SDL_FreeWAV(Uint8 *audio_buf)
 DECLSPEC Uint32 SDLCALL
 SDL_GetWindowFlags(SDL_Window *window)
 {
-    Uint32 flags3 = SDL3_GetWindowFlags(window);
+    Uint32 flags3 = SDL3_GetWindowFlags(window) & SDL_UINT64_C(0x0FFFFFFFF);
```
slouken, line 6164: I think this can just be a cast `(Uint32)`
Most of the flags are compatible, so we just want to truncate to ignore any new flags in the high bits.

sezero: OK, changed to an Uint32 cast

slouken (approved):

PR #174, opened 2024-06-24
Updated for latest SDL3 changes as of ea11c9e0c75b09442c2d2387e866fed2b146ae22

slouken:


sezero (approved):
Attached a tiny patch to sync missing parts in sdl3_include_wrapper.h. OK with that.
[patch.txt](https://github.com/user-attachments/files/15960793/patch.txt)

slouken:
Got it, thanks!

PR #197, opened 2024-07-25
Fix mapping between SDL_EventType in SDL3 to SDL_WindowEventID in SDL2

slouken:
Don't try to be clever, just use a switch statement so they're not tightly coupled.

### DIFF
--- a/src/sdl2_compat.h
+++ b/src/sdl2_compat.h
@@ -648,6 +648,34 @@ typedef struct SDL2_Keysym
 #define SDL2_DISPLAYEVENT 0x150
 #define SDL2_WINDOWEVENT 0x200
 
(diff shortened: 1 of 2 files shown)


src/sdl2_compat.h
```diff
@@ -749,6 +749,33 @@ struct SDL_SysWMinfo
 
 typedef struct SDL_SysWMinfo SDL_SysWMinfo;
 
+typedef enum SDL_WindowEventID
```
sezero, line 752: Maybe move this above, just after `#define SDL2_WINDOWEVENT 0x200`? (Your git seems to be a bit outdated.) And _maybe_ rename to SDL2_WindowEventID for consistency??

slouken: Done!

PR #212, opened 2024-08-31
Updated for the removal of SDL_SCALEMODE_BEST

slouken:


sezero:
What happens if an sdl2 program sets SDL_SCALEMODE_BEST ?

slouken:
I believe they can only do it via the hint, and best has been defined as linear forever. Did I miss something?

sezero:
Um, I actually meant what happens if an sdl2 program uses `SDL_ScaleModeBest` as its `SDL_ScaleMode` parameter in calls such as `SDL_SetTextureScaleMode`

I myself very well may be confused and/or missing something though.

slouken:
Oh, good point. I'll fix that, and add parameter validation in SDL3.

sezero:
OK

PR #217, opened 2024-09-06
Restore log priority after SDL_Quit

madebr:
Doing the following in a SDL project:
```c
    SDL_LogSetPriority(SDL_LOG_CATEGORY_TEST, SDL_LOG_PRIORITY_INFO);
    fprintf(stderr, "before: priority=%d\n", SDL_LogGetPriority(SDL_LOG_CATEGORY_TEST));
    SDL_Quit();
    fprintf(stderr, "after: priority=%d\n", SDL_LogGetPriority(SDL_LOG_CATEGORY_TEST));
```
prints (3, 3) in SDL2:
```
before: priority=3
after: priority=3
```
but prints (3, 1) in SDL3 and current sdl2-compat:
```
before: priority=3
after: priority=1
```

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -5042,6 +5042,8 @@ SDL_InitSubSystem(Uint32 flags)
 SDL_DECLSPEC void SDLCALL
 SDL_Quit(void)
 {
+    int i;
+    SDL_LogPriority priorities[SDL_LOG_CATEGORY_CUSTOM];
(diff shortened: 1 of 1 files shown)


slouken:
Those aren't the default priorities. SDL3 restores the default priorities. It sounds like you want to query priorities before quitting and set those after SDL3_Quit().

slouken (approved):

PR #238, opened 2024-12-31
fix joystick events in WaitEventTimeout, fix HapticRumbleSupported return value

Kethen:
I think SDL_PROP_JOYSTICK_CAP_MONO_LED_BOOLEAN should be mapped to JoystickHasLED true, but I'm unsure

update: not changing JoystickHasLED anymore


src/sdl3_syms.h
```diff
@@ -429,7 +432,7 @@ SDL3_SYM_RENAMED_RETCODE(bool,HapticPause,PauseHaptic,(SDL_Haptic *a),(a),return
 SDL3_SYM_RENAMED_RETCODE(bool,HapticRumbleInit,InitHapticRumble,(SDL_Haptic *a),(a),return)
 SDL3_SYM_RENAMED_RETCODE(bool,HapticRumblePlay,PlayHapticRumble,(SDL_Haptic *a, float b, Uint32 c),(a,b,c),return)
 SDL3_SYM_RENAMED_RETCODE(bool,HapticRumbleStop,StopHapticRumble,(SDL_Haptic *a),(a),return)
-SDL3_SYM_PASSTHROUGH_RETCODE(bool,HapticRumbleSupported,(SDL_Haptic *a),(a),return)
+SDL3_SYM_PASSTHROUGH_INT_BOOL(bool,HapticRumbleSupported,(SDL_Haptic *a),(a),return)
```
sezero, line 435: How did this not error before?

Kethen: [On SDL2, possible return values are -1, SDL_TRUE and SDL_FALSE](https://github.com/libsdl-org/SDL/blob/0efb7c78294b95d62bda35d4cd18945ee578b4d8/src/haptic/SDL_haptic.c#L706C1-L714C2)
[On SDL3, possible return values are true and false](https://github.com/libsdl-org/SDL/blob/0be5b8070d5f1cb501b47f1a5d7c743e5535a883/src/haptic/SDL_haptic.c#L604C1-L611C1)

Kethen: SDL3_SYM_PASSTHROUGH_RETCODE matches type but returns 0 and -1, which flips the outcome

src/sdl2_compat.c
```diff
@@ -8163,7 +8175,9 @@ SDL_JoystickPathForIndex(int idx)
 SDL_DECLSPEC SDL2_bool SDLCALL
 SDL_JoystickHasLED(SDL_Joystick *joystick)
 {
-    return SDL3_GetBooleanProperty(SDL3_GetJoystickProperties(joystick), SDL_PROP_JOYSTICK_CAP_RGB_LED_BOOLEAN, false) ? SDL2_TRUE : SDL2_FALSE;
+    bool has_led = SDL3_GetBooleanProperty(SDL3_GetJoystickProperties(joystick), SDL_PROP_JOYSTICK_CAP_RGB_LED_BOOLEAN, false) ||
+                   SDL3_GetBooleanProperty(SDL3_GetJoystickProperties(joystick), SDL_PROP_JOYSTICK_CAP_MONO_LED_BOOLEAN, false);
+    return has_led ? SDL2_TRUE : SDL2_FALSE;
```
slouken, line 8180: SDL_JoystickHasLED() in SDL2 only returned true for RGB LED support. This change should be dropped.

src/sdl3_syms.h
slouken: This should be changed to a wrapper function rather than adding a new passthrough macro.

src/sdl2_compat.c
Kethen: retracted the commit changing JoystickHasLED

src/sdl3_syms.h
Kethen: Implemented SDL_HapticRumbleSupported as a wrapper function instead

sezero (approved):

sezero:
@slouken: is this good now?

slouken (approved):

slouken:
Merged, thanks!

src/sdl3_syms.h
sezero: > SDL3_SYM_PASSTHROUGH_RETCODE matches type but returns 0 and -1, which flips the outcome

I wonder whether there are any other places where we screwed up like that. (I hope not..)

PR #241, opened 2025-01-06
remove some instances of undefined behavior in the test suite

z-erica:


sezero:
These should first go into SDL2 (and SDL3), then be imported back here.

Then again, let's have a review from maintainers.

icculus:
This seems reasonable.

We can apply this to SDL2 as well. Usually commits flow the other direction, but I guess it's harmless enough.

icculus:
No worries, I'll take care of that.

Okay, I'm merging this one.

icculus:
Hah, SDL3 commented the other one out:

```c
#if 0 /* This test correctly triggers an asan warning: runtime error: signed integer overflow: 2147483647 + 4239 cannot be represented in type 'int' */
    /* Range with min at integer limit */
    min = long_min;
    max = long_max + (Sint32)SDLTest_RandomSint16();
    result = SDLTest_RandomIntegerInRange(min, max);
    SDLTest_AssertPass("Call to SDLTest_RandomIntegerInRange(SINT32_MIN,...)");
    SDLTest_AssertCheck(min <= result && result <= max, "Validated returned value; expected: [%" SDL_PRIs32 ",%" SDL_PRIs32 "], got: %" SDL_PRIs32, min, max, result);
#endif
```

sezero:
So, SDL3 didn't need any update?

icculus:
I guess we can apply the testautomation_sdltest patch, and remove the `#if 0`. The addmul patch isn't necessary in SDL3.

icculus:
Done now!

PR #272, opened 2025-01-31
Add cmake uninstall target

ChillerDragon:
Allows to do the following

```
mkdir build
cd build
cmake ..
sudo make install

sudo make uninstall
```

Same as in https://github.com/libsdl-org/SDL_image/pull/478


madebr (approved):
LGTM, though I feel uncomfortable knowing people will happily run my crappy CMake code as root...

Issue #229, opened 2024-10-08
Generate or provide a non-minimal SDL_config.h for Unix platforms


smcv:
If we want OS distributions (mostly Linux, but also *BSD, Hurd, Haiku, etc.) to eventually replace their SDL2 packages with sdl2-compat, then sdl2-compat's development headers should be as close as is feasible to being a drop-in replacement for "real" SDL2.

Unfortunately, SDL 2 has `SDL_config.h` as part of its API, defining various symbols that are really implementation details of SDL rather than anything that an application should be relying on, such as `HAVE_MEMSET` and `SDL_AUDIO_DRIVER_ALSA`. https://github.com/libsdl-org/sdl12-compat/pull/302, https://github.com/libsdl-org/sdl12-compat/issues/299, https://github.com/libsdl-org/sdl12-compat/issues/297 are examples of places where we needed to make sdl12-compat bug-for-bug compatible with "real" SDL 1.2. I expect that for SDL 2, there should hopefully be less need to do that, but some of it will probably be necessary.

At the moment, in Debian I'm ignoring the SDL_config.h that ships with sdl2-compat, and instead providing [a hand-edited header](https://salsa.debian.org/sdl-team/sdl2-compat/-/blob/debian/latest/debian/SDL_config.h?ref_type=heads) based on the one that Debian's "real SDL2" packaging generates at build-time, which makes some reasonable assumptions about Debian's toolchain, such as:

* we have libc (glibc, in Debian's case) and all Standard C and POSIX headers and functions
* we have all features that are available on any reasonable Unix, like `SDL_TIMER_UNIX`, `SDL_VIDEO_DRIVER_X11` and `HAVE_DBUS_DBUS_H`
* we have Linux-specific functionality like ALSA if and only if `defined(__linux__)` (all official Debian ports are Linux, but there are kFreeBSD and Hurd versions maintained by other-kernel enthusiasts)
* we have Wayland and a modern version of libdecor if and only if `defined(__linux__)` (in principle Wayland should be implementable anywhere, but in Debian's ports, in practice we only have it on Linux)
* we have x86 intrinsics if and only if targeting x86
* we have a basic set of "normal" audio, input, etc. drivers for the platform
* all dlopen'd libraries have their conventional SONAME for the platform (although we could perhaps simplify this into pretending that all libraries are linked with DT_NEEDED rather than dlopen'd, even if in reality they are dlopen'd)

[...]

PR #314, opened 2025-02-04
testautomation: Verify that SIZEOF_VOIDP has been set correctly

smcv:
* testautomation: Verify that SIZEOF_VOIDP has been set correctly
    
    As discussed on #229, if we're going to try to detect this from compiler
    predefined macros like `__LP64__`, we should verify that we've done
    it correctly; otherwise, we could end up incorrectly assuming 32-bit
    pointers on a 64-bit Unix platform that doesn't define `__LP64__`
    (if such a thing exists).
    
    This static assertion succeeds on all Debian architectures where a build
    has been attempted, including all release architectures and several
    unofficial ports. It might fail on non-Linux kernels or non-GNU libc,
    but if it does, our build on those platforms would have been using the
    wrong value for `SIZEOF_VOIDP` already, so it'll only be detecting a
    pre-existing bug.

### DIFF
--- a/test/testautomation_stdlib.c
+++ b/test/testautomation_stdlib.c
@@ -5,7 +5,11 @@
 #include <stdio.h>
 
 #include "SDL.h"
+#include "SDL_config.h"
 #include "SDL_test.h"
+
+/* Compile-time assertions: even better than runtime assertions! */
+SDL_COMPILE_TIME_ASSERT(SIZEOF_VOIDP, SIZEOF_VOIDP == sizeof(void *));
 
 /* Test case functions */
 


sezero:
I am not objecting to the added `_WIN64`assert, but `_LP64` and `__LP64` are predefined macros from the compilier and aren't related to libc headers or kernel: Is there any known example where the existing checks would fail?

sezero:
> I am not objecting to the added `_WIN64`assert, but `_LP64` and `__LP64` are predefined macros

Forgot to add a ref link: https://github.com/cpredef/predef/blob/master/DataModels.md

smcv:
> I am not objecting to the added `_WIN64`assert

Sorry, I'm not sure what added assert you're talking about here?

> but `_LP64` and `__LP64` are predefined macros from the compilier and aren't related to libc headers or kernel: Is there any known example where the existing checks would fail?

Not practical ones, but the goal of this static assertion is to detect whether there are any as-yet unknown examples, and make sure they fail to build, instead of building successfully but then being wrong.

For example if there is somehow a 64-bit Unix platform with the LLP64 data model (int = long = 32-bit, long long = void* = 64-bit, like 64-bit Windows), even if it participates in that convention, it would be correct for it not to define `_LP64` and related macros. We wouldn't want to define `SIZEOF_VOID` to 4 in that case.

Or if there is some obscure compiler (presumably not gcc and not clang) that doesn't define `_LP64` on LP64 platforms, it seems less bad for it to fail to build, rather than building successfully but being wrong.

One example of a platform where this static assertion *would* fail is ARM Morello (CHERI), which has 128-bit pointers. In practice I don't think anyone is playing SDL2 games on these research devices, but, again, better to fail to build than to build wrong.

sezero:
> > I am not objecting to the added `_WIN64`assert
> 
> Sorry, I'm not sure what added assert you're talking about here?

That was my touchpad playing mischief :(
Edited my post above to correct it

> One example of a platform where this static assertion _would_ fail is ARM Morello (CHERI), which has 128-bit pointers.

OK

smcv:
> One example of a platform where this static assertion _would_ fail is ARM Morello (CHERI), which has 128-bit pointers

In fact I believe the last commit I added to #315 should fix that, assuming that nobody is trying to use a non-gcc non-clang compiler for such a machine.

PR #315, opened 2025-02-04
Improve SDL_config_unix.h

smcv:
* SDL_config.h: Assume that Unix platforms define `__unix__`
    
    This is believed to be a predefined macro on all reasonable Unix
    platforms in practice (it's certainly predefined on GNU/Linux).
    We can add some `|| defined(__myplatform__)` as necessary, but
    this way is more readable than listing every Unix we know about
    individually.
    
    https://sourceforge.net/p/predef/wiki/OperatingSystems/ notes that
    "the xlC or the DEC C/C++ compiler" don't define these, but that doesn't
    seem likely to have any practical impact on us.

* SDL_config.h: Add a #warning if we don't recognise the platform
    
    SDL_config_minimal.h is rather minimal, and we probably don't want anyone
    to actually be relying on it.

* SDL_config_unix.h: Describe our basis for assuming that headers exist
    
    Some of these are Standard C, some are POSIX, and a few are non-standard
    but widely available.

* SDL_config_unix.h: Only define HAVE_GCC_ATOMICS for gcc or clang
    
    The gcc atomic operations are a gcc-specific extension, which clang
    implements for compatibility. In the unlikely event that someone compiles
    sdl2-compat with some other compiler, we can't assume their presence.

* SDL_config_unix.h: Try to use `__SIZEOF_POINTER__`, if defined
    
    gcc (and presumably clang) predefines this, and if present, it seems
    reasonable to treat it as authoritative.

---

As discussed in #229. Compiles successfully on my Debian system, otherwise untested.

### DIFF
--- a/include/SDL2/SDL_config.h
+++ b/include/SDL2/SDL_config.h
@@ -51,6 +51,7 @@
 #include "SDL_config_unix.h"
 #else
 /* This is a minimal configuration just to get SDL running on new platforms. */
+#warning SDL_config.h might be incomplete, good luck
 #include "SDL_config_minimal.h"
 #endif /* platform config */
 
--- a/include/SDL2/SDL_config.h
+++ b/include/SDL2/SDL_config.h
@@ -47,7 +47,7 @@
 #include "SDL_config_emscripten.h"
 #elif defined(__NGAGE__)
 #include "SDL_config_ngage.h"
-#elif defined (__LINUX__) || defined(__FREEBSD__) || defined(__NETBSD__) || defined(__OPENBSD__) || defined(__SOLARIS__)
+#elif defined(__unix__)
 #include "SDL_config_unix.h"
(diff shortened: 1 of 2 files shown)


include/SDL2/SDL_config.h
```diff
@@ -51,6 +51,7 @@
 #include "SDL_config_unix.h"
 #else
 /* This is a minimal configuration just to get SDL running on new platforms. */
+#warning SDL_config.h might be incomplete, good luck
```
sezero, line 54: Is `#warning` directive universally supported? A `#pragma message("...")` may also be considered here.

slouken: I believe `#warning` is more widely supported than `#pragma message()`

slouken:
Merged, thanks!

PR #318, opened 2025-02-05
Expand SDL_config_unix.h

smcv:
* SDL_config_unix: Assume that Unix platforms build SDL with libc
    
    On platforms like Windows it's reasonable to rely on a platform-specific
    interface like Win32 for I/O, but on Unix, libc stdio is more likely to
    be the only thing there is.

* SDL_config_unix.h: Assume we have a complete set of Standard C functions
    
    Classic SDL2 checked for these functions, and defined the corresponding
    macros if found. According to the man pages on my Linux system, these
    are all required by C11 or older; in 2025, using a C11 compiler and
    standard library doesn't seem like too much to ask.

* SDL_config_unix.h: Assume we have complete POSIX functionality
    
    Classic SDL2 checked for these functions at compile-time and defined
    the corresponding macros if found. According to the man pages on my
    Linux system, these were all specified in POSIX.1-2008 or older, so
    they seem like they ought to be a safe assumption.

* SDL_config_unix.h: Assume a few widely-implemented non-standard functions
    
    bcopy(), index() and rindex() are no longer specified by POSIX, but they
    were in older POSIX standards, so any platform with a stable ABI cannot
    actually remove them.
    
    alloca() is widely-implemented, and we already assumed that `<alloca.h>`
    exists.

* SDL_config_unix.h: Define macros for a few GNU extensions if using glibc
    
    glibc provides getauxval(), strcasestr() and pthread_setname_np(),
    and recent versions also provide memfd_create(). Linux games could
    conceivably be relying on classic SDL2 checking for these.
    
    I'm checking for `__GLIBC__` rather than `__LINUX__` because these are
    really features of the libc, not the kernel.

* SDL_config_unix.h: Assume all Unix platforms have POSIX functionality
    
    To the best of my knowledge these SDL features are based on POSIX
    function calls, so any platform that calls itself Unix should be able
    to provide them.

* SDL_config_unix.h: Assume that Unix platforms have GLX
    
    This is a de facto standard on Unix, and we're already assuming that
    such platforms have X11.

* SDL_config_unix.h: Assume that Linux has EGL

* SDL_config_unix.h: Assume typical functionality on Linux
    
    Linux games could conceivably be relying on these feature-tests.

* SDL_config_unix.h: Assume that Linux has XInput2

    Classic SDL2 defined macros for several X11 extensions, but the only one
    for which I found a concrete example in codesearch.debian.net is this one,
    which the glvis package relies on.
    
    I don't know whether other Unix platforms can be relied on to have this,
    but it seems like a safe assumption on Linux.

---

Based on reviewing classic SDL2's generated `SDL_config.h` on my Linux system, and comparing the macros it defines with the corresponding functions' man pages and my own knowledge of Unix portability runes. These can all be tweaked as necessary if we find that there is a Unix platform that doesn't meet these assumptions, but it seems maybe safer to mirror classic SDL's macros than to only provide a small subset?


smcv:
Some macros defined by my `SDL_config.h` that are not accounted for here:

* `HAVE_WCSLCPY`
* `HAVE_WCSLCAT`
* `HAVE_STRLCPY`
* `HAVE_STRLCAT`
* `HAVE_SA_SIGACTION`
* `HAVE_FCITX`
* `HAVE_IBUS_IBUS_H`
* `HAVE_LIBSAMPLERATE_H`

slouken:
Merged, thanks!

Issue #309, opened 2025-02-03
Crashes while calling SDL_FreeSurface


peterix:
Hello,

Recently, Arch switched to SDL3 + SDL2-compat and this has caused some issues.
I use SDL2 for a zig port of a platformer. Code here: https://github.com/peterix/OpenTitus

I load some pallette based sprites (with bit planes and other funky things) from original game data, and then use a cache for surfaces in the window pixel format.

The surfaces that crash are created here:
https://github.com/peterix/OpenTitus/blob/020c24a7b9f17ab02d74f4455f570dbad694686a/src/sprites.zig#L247

And the crash happens here:
https://github.com/peterix/OpenTitus/blob/020c24a7b9f17ab02d74f4455f570dbad694686a/src/sprites.zig#L266

Stack trace:
```
> lldb ./opentitus 
(lldb) target create "./opentitus"
Current executable set to '/home/peterix/projects/OpenTitus/bin/titus/opentitus' (x86_64).
(lldb) r
Process 89270 launched: '/home/peterix/projects/OpenTitus/bin/titus/opentitus' (x86_64)
Process 89270 stopped
* thread #1, name = 'opentitus', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x19)
    frame #0: 0x00007ffff7cd58d8 libc.so.6`__GI___libc_free(mem=0x0000000000000021) at malloc.c:3453:7
(lldb) bt
* thread #1, name = 'opentitus', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x19)
  * frame #0: 0x00007ffff7cd58d8 libc.so.6`__GI___libc_free(mem=0x0000000000000021) at malloc.c:3453:7
    frame #1: 0x00007ffff7919a69 libSDL3.so.0`SDL_DestroySurface_REAL [inlined] SDL_free_REAL(ptr=<unavailable>) at SDL_malloc.c:6505:5
    frame #2: 0x00007ffff7919a63 libSDL3.so.0`SDL_DestroySurface_REAL [inlined] SDL_DestroySurface_REAL(surface=0x000000000265ae90) at SDL_surface.c:2978:9
    frame #3: 0x00007ffff79199eb libSDL3.so.0`SDL_DestroySurface_REAL(surface=0x000000000265ae90) at SDL_surface.c:2943:6
    frame #4: 0x000000000144977a opentitus`src.SDL.freeSurface(surface=0x00000000029740f0) at SDL.zig:113:26
[...]

PR #323, opened 2025-02-06
init: Don't call into SDL3 _at all_ during dllinit().

icculus:
This uses platform-specific code for some tasks (get/set environment, logging, etc), implements some simple things like strlen, and moves SDL2Compat_InitOnStartup to SDL_InitSubSystem and wraps it in an SDL_InitState.

This removes some questions of SDL3's intialization status during these early moments and, more importantly, makes sure we don't call into SDL_malloc before the app has a chance to set their own allocators.

Reference Issue #309.


icculus:
Note that SDL2 and SDL3 should probably pull in the dynapi getenv change, too, since even before SDL3 made SDL_getenv allocate a bunch of memory, the Windows implementation would still call SDL_realloc.

sdl12-compat might want to steal some of this, too, so it doesn't call into an SDL2 that allocates memory either.

slouken (approved):

sezero:
Is `@autoreleasepool` needed in `SDL2Compat_NSLog`? Why? (Forgive my objc ignorance here..)

icculus:
I'm not actually sure it's needed, but I copied that function from SDL3, and it's harmless to be there even if it turns out to be unnecessary.

sezero:
> I'm not actually sure it's needed, but I copied that function from SDL3, and it's harmless to be there even if it turns out to be unnecessary.

Is this equivalent / correct?

```objc
    NSString *nsText = [NSString stringWithCString:text encoding:NSUTF8StringEncoding];
    if (prefix && *prefix) {
        NSString *nsPrefix = [NSString stringWithCString:prefix encoding:NSUTF8StringEncoding];
        NSLog(@"%@%@", nsPrefix, nsText);
    } else {
        NSLog(@"%@", nsText);
    }
```

slouken:
> Is this equivalent / correct?
> 
> ```objc
>     NSString *nsText = [NSString stringWithCString:text encoding:NSUTF8StringEncoding];
>     if (prefix && *prefix) {
>         NSString *nsPrefix = [NSString stringWithCString:prefix encoding:NSUTF8StringEncoding];
>         NSLog(@"%@%@", nsPrefix, nsText);
>     } else {
>         NSLog(@"%@", nsText);
>     }
> ```

This seems correct to me, but you still need it in an `@autoreleasepool` because you're allocating an NSString. Would you like to make an SDL3 PR for this change?

sezero:
>  you still need it in an `@autoreleasepool` because you're allocating an NSString. Would you like to make an SDL3 PR 

No need then -- I was just trying to get rid of `@autoreleasepool` :)

slouken:
> > you still need it in an `@autoreleasepool` because you're allocating an NSString. Would you like to make an SDL3 PR
> 
> No need then -- I was just trying to get rid of `@autoreleasepool` :)

Oh, okay. Nice try. :)

This does seem like a good change in general though, I'll throw it into a bug so we can look at it later.

sezero:
> Oh, okay. Nice try. :)

How about this then? (w/o whitespace changes)

```diff
--- sdl2_compat_objc.m~
+++ sdl2_compat_objc.m
@@ -37,13 +37,13 @@
 SDL2_PRIVATE void SDL2Compat_NSLog(const char *prefix, const char *text)
 {
-    @autoreleasepool {
-        NSString *nsText = [NSString stringWithUTF8String:text];
+        NSString *nsText = [[NSString alloc] initWithUTF8String:text];
         if (prefix && *prefix) {
-            NSString *nsPrefix = [NSString stringWithUTF8String:prefix];
+            NSString *nsPrefix = [[NSString alloc] initWithUTF8String:prefix];
             NSLog(@"%@%@", nsPrefix, nsText);
+            [nsPrefix release];
         } else {
             NSLog(@"%@", nsText);
         }
+        [nsText release];
-    }
 }
 
```

slouken:
You can't guarantee that NSString doesn't do internal allocations. All calls into Cocoa APIs should have a toplevel autorelease pool to prevent leaks.

sezero:
It was the way glib does it, but I see, OK. Not going after this further.

Issue #329, opened 2025-02-08
Distorted audio with moonlight-qt and sdl2-compat on Arch Linux


rubin55:
I'm running Arch Linux. I use `moonlight-qt` to stream my desktop (which runs `sunshine`) to my laptop (all systems running Arch Linux).

Since the upgrade of `sdl3` and introduction of `sdl2-compat`, I have highly distorted audio. It is specifically not due to network latency, which sounds very different. This sounds like way too loud oversteering or clipping, robotic. It sounds similar to libsdl-org/SDL#12150.

I've made a video to illustrate the issue (zipped, because I can apparently not upload an mp4 directly):
[moonlight-qt-sdl2-compat-distorted-audio.zip](https://github.com/user-attachments/files/18721749/moonlight-qt-sdl2-compat-distorted-audio.zip)

slouken:
Thanks, we'll take a look!

cgutman:
Interesting, I was actually just running this test with the exact same configuration (Arch Linux too) and audio was okay outputing to an HDMI audio device (monitor).

When you start a stream Moonlight will print some message about the audio configuration which look like:
```
00:00:06 - SDL Warn (0): Desired audio buffer: 720 samples (5760 bytes)
00:00:06 - SDL Warn (0): Obtained audio buffer: 720 samples (5760 bytes)
00:00:06 - SDL Warn (0): SDL audio driver: pulseaudio
00:00:06 - SDL Warn (0): Audio channel count: 2
00:00:06 - SDL Warn (0): Audio channel mask: 3
```

Can you post those from your test?

rubin55:
In my case its:

```
00:00:09 - SDL Warn (0): Starting audio stream...
00:00:09 - SDL Warn (0): Desired audio buffer: 720 samples (5760 bytes)
00:00:09 - SDL Warn (0): Obtained audio buffer: 720 samples (5760 bytes)
00:00:09 - SDL Warn (0): SDL audio driver: pipewire
00:00:09 - SDL Warn (0): Received first video packet after 0 ms
00:00:09 - SDL Warn (0): Audio stream has 2 channels
```

rubin55:
Setting the audio driver to `pulseaudio` gets rid of the distortion when using `sdl2-compat`:

```
export SDL_AUDIODRIVER=pulseaudio
moonlight
```

slouken:
[...]

PR #330, opened 2025-02-08
logging: Wrap log output function to fixup log priority values

cgutman:
I noticed in #329 that all of Moonlight's info logs were appearing as warnings when running under sdl2-compat and tracked it down to this bug.

One of the parameters to the `SDL_LogOutputFunction` is `SDL_LogPriority`, which changed values between SDL2 and SDL3, so we can't just pass-through the caller's logging function as-is. We need to convert the priority enum value from SDL3 to SDL2 before each call to the log output function.

The catch here is that the caller could have saved the original `SDL_LogOutputFunction` (from SDL3) then tried to restore it later. Since performing the 3to2 conversion would confuse SDL3's default logger, we specifically check for this case and skip the 3to2 conversion thunk if it's detected.


src/sdl2_compat.c
```diff
@@ -1333,6 +1334,31 @@ SDL_LOG_IMPL(Error, ERROR)
...
+    } else {
+        LogOutputFunction2 = callback;
+        SDL3_SetLogOutputFunction(LogOutputFunction3to2, userdata);
+    }
+}
+
+SDL_DECLSPEC void SDLCALL SDL_LogGetOutputFunction(SDL2_LogOutputFunction *callback, void **userdata)
+{
+    SDL3_GetLogOutputFunction((SDL_LogOutputFunction *)callback, userdata);
+    if (callback && (SDL_LogOutputFunction)*callback == LogOutputFunction3to2) {
```
cgutman, line 1356: It's not explicitly documented that you can pass NULL for one of these parameters, but real SDL2 lets you so I handle that too.

src/sdl2_compat.c
```diff
@@ -1333,6 +1334,31 @@ SDL_LOG_IMPL(Error, ERROR)
...
 #undef SDL_LOG_IMPL
 
+static void SDLCALL LogOutputFunction3to2(void *userdata, int category, SDL_LogPriority priority, const char *message)
+{
+    LogOutputFunction2(userdata, category, LogPriority3to2(priority), message);
+}
+
+SDL_DECLSPEC void SDLCALL SDL_LogSetOutputFunction(SDL2_LogOutputFunction callback, void *userdata)
+{
+    if ((SDL_LogOutputFunction)callback == SDL3_GetDefaultLogOutputFunction()) {
```
cgutman, line 1344: Hmm, looks like you can pass a NULL callback too to disable log output, I better handle that as well...

sezero (approved):

slouken:
Merged, thanks!

PR #332, opened 2025-02-09
Fix joystick device list before polling event

jpy794:
[sdl-jstest](https://github.com/Grumbel/sdl-jstest/blob/60f6e62109e100b6ef363841b71c2e37ed714d01/src/sdl2-jstest.c#L145) would call `SDL_JoystickOpen()` before polling SDL event. This breaks joystick device list in sdl2-compat and always gets null returned.
Initializing device list in `SDL_init()` would fix it.


slouken:
Looks good, thanks!

PR #339, opened 2025-02-09
env: sync environment variables during subsystem init

cgutman:
SDL3 uses a cached environment block that is managed internally and isn't impacted by applications calling `setenv()` on their own. This is a behavior change from SDL2. Moonlight [uses setenv](https://github.com/moonlight-stream/moonlight-qt/blob/bdd9a3a9947ce621730187dc4dc66b5f935bf00d/app/main.cpp#L643-L691) to adjust various non-hints which aren't taking effect due to this bug. One result of this is that we no longer use the native Wayland backend under sdl2-compat like we did under real SDL2.

I'm not sure if there are better options to deal with this behavior change here (like an API in SDL3 to tell it to refresh the cached environment), but for now I've written a simple `SynchronizeEnvironmentVariable()` that I sprinkled calls to in each place where we call `SDL_InitSubsystem()`. This is sufficient to fix Moonlight's use case of these environment variables.


src/sdl2_compat.c
```diff
@@ -5567,10 +5567,30 @@ SDL_RemoveTimer(SDL2_TimerID id)
...
+        SDL3_free(fresh_envp);
+    }
+
+    SDL3_DestroyEnvironment(fresh_env);
+}
 
 SDL_DECLSPEC int SDLCALL
 SDL_AudioInit(const char *driver_name)
 {
+    SynchronizeEnvironmentVariables();
```
slouken, line 5593:
```suggestion
    SynchronizeEnvironmentVariables();

```

src/sdl2_compat.c
```diff
@@ -5813,6 +5836,8 @@ SDL_InitSubSystem(Uint32 flags)
         return -1;
     }
 
+    SynchronizeEnvironmentVariables();
```
slouken, line 5839: I don't know if it's a big deal, but this is going to synchronize multiple times if flags is SDL_INIT_VIDEO or SDL_INIT_AUDIO

slouken:
Nice catch, thanks!

PR #343, opened 2025-02-10
Fix scaling for legacy programs on wayland

simonioc:
I think this is the solution reached after https://github.com/libsdl-org/SDL/issues/12158
Quirks will eventually be added to applications that do not behave well with this

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1070,6 +1070,7 @@ SDL2Compat_InitOnStartupInternal(void)
     SDL3_SetHint("SDL_WINDOWS_DPI_AWARENESS", "unaware");
     SDL3_SetHint("SDL_BORDERLESS_WINDOWED_STYLE", "0");
     SDL3_SetHint("SDL_VIDEO_SYNC_WINDOW_OPERATIONS", "1");
+    SDL3_SetHint("SDL_VIDEO_WAYLAND_SCALE_TO_DISPLAY","1");
 
     SDL2Compat_InitLogPrefixes();
 
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1070,7 +1070,11 @@ SDL2Compat_InitOnStartupInternal(void)
     SDL3_SetHint("SDL_WINDOWS_DPI_AWARENESS", "unaware");
     SDL3_SetHint("SDL_BORDERLESS_WINDOWED_STYLE", "0");
     SDL3_SetHint("SDL_VIDEO_SYNC_WINDOW_OPERATIONS", "1");
-    SDL3_SetHint("SDL_VIDEO_WAYLAND_SCALE_TO_DISPLAY","1");
+
+    // Pretend Wayland doesn't have fractional scaling
+    // This is more compatible with applications that have only been tested under X11 without high DPI support
+    // Full discussion is here: https://github.com/libsdl-org/SDL/issues/12158
+    SDL3_SetHint(SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "1");
 
     SDL2Compat_InitLogPrefixes();
 


cgutman:
What does SDL2's Wayland backend do if apps that are already using it?

If we're just concerned about behavior changes for apps previously defaulting to X11, we should avoid setting this option if the app is already requesting Wayland via `SDL_VideoInit()` or `SDL_VIDEODRIVER`. Those apps would have already been tested with the SDL2 Wayland backend and be expecting the high DPI behavior it had.

src/sdl2_compat.c
```diff
@@ -1070,6 +1070,7 @@ SDL2Compat_InitOnStartupInternal(void)
     SDL3_SetHint("SDL_WINDOWS_DPI_AWARENESS", "unaware");
     SDL3_SetHint("SDL_BORDERLESS_WINDOWED_STYLE", "0");
     SDL3_SetHint("SDL_VIDEO_SYNC_WINDOW_OPERATIONS", "1");
+    SDL3_SetHint("SDL_VIDEO_WAYLAND_SCALE_TO_DISPLAY","1");
```
slouken, line 1073:
```suggestion
    SDL3_SetHint(SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "1");
```

slouken: Also it's probably worth adding a comment describing the effect of this so people reading the code later understand why it was added and when they should set the quirk to turn this off.

slouken:
> What does SDL2's Wayland backend do for apps that are already using it?
> 
> If we're just concerned about behavior changes for apps previously defaulting to X11, we should avoid setting this option if the app is already requesting Wayland via `SDL_VideoInit()` or `SDL_VIDEODRIVER`. Those apps would have already been tested with the SDL2 Wayland backend and be expecting the high DPI behavior it had.

Agreed.

simonioc:
> What does SDL2's Wayland backend do for apps that are already using it?
> 
> If we're just concerned about behavior changes for apps previously defaulting to X11, we should avoid setting this option if the app is already requesting Wayland via `SDL_VideoInit()` or `SDL_VIDEODRIVER`. Those apps would have already been tested with the SDL2 Wayland backend and be expecting the high DPI behavior it had.

I thought about that but i'm not sure whether SDL distinguishes the hints set by the app and the ones set by the user through env vars.

slouken:
@cgutman, you want to take a stab at this? You're more familiar with getting these details right.

cgutman:
Yeah, I can give it a shot.

However, I'm not completely certain that I understand the expected behavior in all cases. Specifically, I'm wondering how `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY` interacts with `SDL_WINDOW_HIGH_PIXEL_DENSITY`.

Maybe @Kontrabant can chime in what we want `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY` to be in the following cases to best match SDL2 behavior:
- `SDL_VIDEODRIVER=wayland` + `SDL_WINDOW_ALLOW_HIGHDPI` set
- `SDL_VIDEODRIVER=x11/unset` + `SDL_WINDOW_ALLOW_HIGHDPI` set
- `SDL_VIDEODRIVER=wayland` + `SDL_WINDOW_ALLOW_HIGHDPI` unset
- `SDL_VIDEODRIVER=x11/unset` + `SDL_WINDOW_ALLOW_HIGHDPI` unset

slouken:
> Yeah, I can give it a shot.
> 
> However, I'm not completely certain that I understand the expected behavior in all cases. Specifically, I'm wondering how `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY` interacts with `SDL_WINDOW_HIGH_PIXEL_DENSITY`.

Unfortunately `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY` needs to be set when video initialization happens, so we don't know in advance whether the application sets `SDL_WINDOW_ALLOW_HIGHDPI`. That's where the quirk system will come in, we'll need to set the environment variable to "0" for applications that set that flag (and leave the default video driver).

We'll also need to be able to distinguish between the user setting the environment variable to force wayland and the application doing it because they've tested fractional scaling under Wayland.

My hypothesis is that the number of SDL2 applications that handle high DPI properly is very low and we can just blanket set the hint and add quirks for the few that do.

Kontrabant:
> However, I'm not completely certain that I understand the expected behavior in all cases. Specifically, I'm wondering how SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY interacts with SDL_WINDOW_HIGH_PIXEL_DENSITY.

One of the unfortunate quirks of scale-to-display is that it universally overrides high pixel density, because they can't really coexist. The former forces everything, from window to display dimensions, to appear as pixels, simulating a Win32/X11 style environment. I tried allowing them to coexist at one point and decided against it, as it was simply too messy.

slouken:
I'm going to go ahead and merge this so we can get feedback. @cgutman, feel free to follow up with a more refined PR.

PR #344, opened 2025-02-11
Scale to display only for apps previously using X11 and fix some init bugs

cgutman:
This is a continuation of #343 to implement the (hopefully) more accurate logic of guessing that the app wants `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY=0` if it asks for Wayland specifically and `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY=1` if not.

To avoid additional code duplication and fix some pre-existing bugs, I also moved all the subsystem-specific pre and post initialization steps to a separate function. There was some compatibility code that was only running if the app used `SDL_InitSubsystem()` rather than `SDL_VideoInit()` or `SDL_AudioInit()`.

This PR is easier to review commit-by-commit.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1071,11 +1071,6 @@ SDL2Compat_InitOnStartupInternal(void)
     SDL3_SetHint("SDL_BORDERLESS_WINDOWED_STYLE", "0");
     SDL3_SetHint("SDL_VIDEO_SYNC_WINDOW_OPERATIONS", "1");
 
-    // Pretend Wayland doesn't have fractional scaling
-    // This is more compatible with applications that have only been tested under X11 without high DPI support
-    // Full discussion is here: https://github.com/libsdl-org/SDL/issues/12158
-    SDL3_SetHint(SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "1");
-
     SDL2Compat_InitLogPrefixes();
 
     return true;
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -5593,15 +5593,94 @@ static void SynchronizeEnvironmentVariables()
     SDL3_DestroyEnvironment(fresh_env);
 }
 
+static SDL_InitFlags PreInitSubsystem(SDL_InitFlags flags)
+{
+    /* If the subsystem is already initialized, mask out the flag for it */
+    flags &= ~SDL_WasInit(flags);
```
slouken, line 5594: This isn't quite correct. This will prevent SDL_InitSubSystem(SDL_INIT_VIDEO) twice from incrementing the refcount twice and the first SDL_QuitSubSystem(SDL_INIT_VIDEO) will shut down the video subsystem.

cgutman: This masked flag value is only used for determining which subsystem pre/post initialization steps we need to perfom (only when the subsystem goes from refcount 0->1). We're still passing the caller's unmodified flags value to `SDL3_InitSubsystem()`, so the subsystem refcount should still be accurate.

slouken: Ah, okay.

src/sdl2_compat.c
```diff
@@ -5619,6 +5614,22 @@ static SDL_InitFlags PreInitSubsystem(SDL_InitFlags flags)
...
+        if (!hint) {
+            hint = SDL3_GetHint("SDL_VIDEODRIVER");
+        }
+        if (!hint || SDL3_strcasecmp(hint, "wayland") != 0) {
+            /* If this app does not explicitly ask for Wayland, avoid using Wayland fractional scaling.
+             * This is more compatible with applications that have only been tested under X11 without high DPI support.
+             * Full discussion is here: https://github.com/libsdl-org/SDL/issues/12158 */
+            SDL3_SetHint(SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "1");
+        } else {
+            /* If the app did ask for Wayland, let's assume it can handle the standard Wayland behavior in SDL3 */
```
slouken, line 5630: This will also get hit if the user set SDL_VIDEODRIVER=wayland in the environment. In this case we definitely want to continue to set `SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY`

cgutman: Okay, this makes the problem a lot more complicated. The big problem is trying to distinguish between a user setting SDL_VIDEODRIVER=wayland and the app internally doing it.

The only way I can think to do it is to snapshot the environment with `SDL_CreateEnvironment(true)` in `SDL2Compat_InitOnStartupInternal()`, unset the variable in the real environment, and then see if we observe it set again later on. The problem with this approach is that unsetting `SDL_VIDEODRIVER` is observable by the app and could break its logic.

Maybe a quirk is best after all...

slouken: In any case the cleanup seems like a good change.

cgutman: I put up a second PR #346 with just the init refactoring (and a little more than I did here)

cgutman:
Closing as superseded by #345 and #346.

PR #350, opened 2025-02-12
Implement several hints changed/removed in SDL3

cgutman:
This PR implements:
- `SDL_HINT_VIDEO_HIGHDPI_DISABLED`
- `SDL_HINT_MOUSE_RELATIVE_SCALING`
- `SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4` (inverted from SDL3)

I also removed `SDL2Compat_GetHint()` and `SDL2Compat_GetHintBoolean()` and replaced them with the SDL3 functions directly. `SDL2Compat_GetHint()` was only looking at environment variables, so apps that called `SDL_SetHint()` to adjust `SDL_HINT_RENDER_BATCHING` would not have those changes take effect. `SDL2Compat_GetHintBoolean()` also did not properly handle the hint value `"true"` (returning false in that case).

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -500,25 +500,6 @@ SDL2Compat_GetExeName(void)
     return exename;
 }
 
-static const char *
-SDL2Compat_GetHint(const char *name)
-{
-    return SDL3_getenv(name);
-}
-
-static bool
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #352, opened 2025-02-13
Added SDL2COMPAT_ASAN build option

slouken:


CMakeLists.txt
```diff
@@ -94,6 +95,75 @@ if(SDL2COMPAT_STATIC AND NOT (CMAKE_SYSTEM_NAME MATCHES "Linux"))
...
+  endif()
+
+  if(CMAKE_CXX_COMPILER)
+    check_cxx_compiler_flag (${FLAG} HAS_CXX_FLAG_${ASAN_FLAG})
+    if (HAS_CXX_${ASAN_FLAG})
+      string(APPEND CMAKE_CXX_FLAGS_DEBUG " ${FLAG}")
+    endif()
+  endif()
+
+  cmake_pop_check_state()
```
madebr, line 140:
```suggestion
  endif()

  cmake_pop_check_state()
```

CMakeLists.txt
```diff
@@ -94,6 +95,75 @@ if(SDL2COMPAT_STATIC AND NOT (CMAKE_SYSTEM_NAME MATCHES "Linux"))
...
+
+  if(CMAKE_CXX_COMPILER)
+    check_cxx_compiler_flag(${FLAG} HAS_CXX_${SUFFIX})
+    if(HAS_CXX_${SUFFIX})
+      string(APPEND CMAKE_CXX_FLAGS_DEBUG " ${FLAG}")
+    endif()
+  endif()
+endmacro()
+
+macro(asan_check_add_debug_flag ASAN_FLAG)
```
madebr, line 112:
```suggestion
  endif()

macro(asan_check_add_debug_flag ASAN_FLAG)
```

madebr:
I found the issue:
I don't install SDL3 system-wide, but rely on the runpath in sdl2-compat's `libSDL2-2.0.so.0` to find `libSDL3.so.0`.
It looks like this path is not taken into account when using `dlopen` from my libasan.
If I run with the `LD_LIBRARY_PATH` environment variable set, then everything works as expected (with asan enabled).

icculus:
This breaks ffplay again, sorry.  :(

Just run "ffplay" with any image or movie file to trigger it.

icculus:
(the merged version had fixes for ffplay, which works again, for those keeping track.)

src/sdl2_compat.c
```diff
@@ -456,19 +456,62 @@ static QuirkEntryType quirks[] = {
...
+                        while (sep > spot) {
+                            --sep;
+                            if (*sep == '/') {
+                                ++sep;
+                                break;
+                            }
+                        }
+                        spot = sep;
+                    }
+                    strlcpy(buf, spot, maxpath);
```
sezero, line 505: `strlcpy` isn't available everywhere, only recent glibc has it as I remember. Fails build for me, for e.g.

sezero: Should I change it to `snprint(buf, maxpath, "%s", spot);` ?

slouken: Can we use SDL3 functions at this point? If not, then yes please!

sezero: > Can we use SDL3 functions at this point? If not, then yes please!

Not 100% sure, so I pushed the snprintf patch

That linux procedure uses several other libc calls which have SDL counterparts, strchr for e.g., so if you want to change to SDL calls, it won't be just for strlcpy.

slouken: It looks like probably not ;)
```
/* DO NOT USE SDL3 FUNCTIONS IN HERE! */
```

icculus: SDL2COMPAT_stpcpy() if we're still in dllinit, please.

slouken: I'll fix that. Also rename it to SDL2COMPAT_strlcpy so it's more familiar with someone coming from SDL3 ;)

slouken: er, or not. :)

icculus: yeah, different interface.  :)

sezero: > SDL2COMPAT_stpcpy() if we're still in dllinit, please.

Areyou sure? There is no buffer size check in that

sezero: Also there are other libc calls in there, e.g str[r]chr, strstr, etc...

sezero: > yeah, different interface. :)

Yup: https://linux.die.net/man/3/stpcpy

slouken: Okay, let’s leave it as-is for now. We have bigger fish to fry 😊

PR #361, opened 2025-02-15
hints: Handle envvars that turned into hints in SDL3

cgutman:
Some of the environment variables in SDL2 have been removed and replaced with hints in SDL3. This PR covers everything mentioned in `README-migration.md` except variables that changed semantics in SDL3 (`SDL_DISKAUDIODELAY`, `SDL_DUMMYAUDIODELAY`, and `SDL_AUDIO_ALSA_DEBUG`).

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -5827,7 +5827,7 @@ SDL_AudioInit(const char *driver_name)
     SynchronizeEnvironmentVariables();
 
     if (driver_name) {
-        SDL3_SetHint(SDL_HINT_AUDIO_DRIVER, driver_name);
+        SDL3_SetHintWithPriority(SDL_HINT_AUDIO_DRIVER, driver_name, SDL_HINT_OVERRIDE);
     }
 
     return InitSubsystemInternal(SDL_INIT_AUDIO);
@@ -6035,7 +6035,7 @@ SDL_VideoInit(const char *driver_name)
     SynchronizeEnvironmentVariables();
 
     if (driver_name) {
-        SDL3_SetHint(SDL_HINT_VIDEO_DRIVER, driver_name);
+        SDL3_SetHintWithPriority(SDL_HINT_VIDEO_DRIVER, driver_name, SDL_HINT_OVERRIDE);
     }
 
     return InitSubsystemInternal(SDL_INIT_VIDEO);
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -5624,6 +5624,79 @@ SDL_RemoveTimer(SDL2_TimerID id)
     return SDL3_RemoveTimer((SDL_TimerID)id) ? SDL2_TRUE : SDL2_FALSE;
 }
 
+typedef enum {
+    V2H_PASSTHROUGH,   /* Leave value unchanged */
+    V2H_BOOL,          /* 0 = value unset, 1 = value set */
+    V2H_BOOL_INVERTED, /* 1 = value unset, 0 = value set */
+} VarToHintConversionType;
+
+static struct {
+    const char *old_envvar;
+    const char *new_hint;
+    VarToHintConversionType type;
+} envvars_to_hints[] = {
+    { "SDL_DISKAUDIOFILE", SDL_HINT_AUDIO_DISK_OUTPUT_FILE, V2H_PASSTHROUGH },
+    { "SDL_DISKAUDIOFILEIN", SDL_HINT_AUDIO_DISK_INPUT_FILE, V2H_PASSTHROUGH },
+    { "SDL_HIDAPI_DISABLE_LIBUSB", SDL_HINT_HIDAPI_LIBUSB, V2H_BOOL_INVERTED },
+    { "SDL_HIDAPI_JOYSTICK_DISABLE_UDEV", SDL_HINT_HIDAPI_UDEV, V2H_BOOL_INVERTED },
+#ifdef SDL_PLATFORM_FREEBSD
+    { "SDL_INPUT_FREEBSD_KEEP_KBD", SDL_HINT_MUTE_CONSOLE_KEYBOARD, V2H_BOOL_INVERTED },
+#endif
+#ifdef SDL_PLATFORM_LINUX
+    { "SDL_INPUT_LINUX_KEEP_KBD", SDL_HINT_MUTE_CONSOLE_KEYBOARD, V2H_BOOL_INVERTED },
+#endif
+#ifdef SDL_PLATFORM_VITA
+    { "VITA_DISABLE_TOUCH_BACK", SDL_HINT_VITA_ENABLE_BACK_TOUCH, V2H_BOOL_INVERTED },
+    { "VITA_DISABLE_TOUCH_FRONT", SDL_HINT_VITA_ENABLE_FRONT_TOUCH, V2H_BOOL_INVERTED },
+    { "VITA_MODULE_PATH", SDL_HINT_VITA_MODULE_PATH, V2H_PASSTHROUGH },
+    { "VITA_PVR_OGL", SDL_HINT_VITA_PVR_OPENGL, V2H_BOOL },
+    { "VITA_PVR_SKIP_INIT", SDL_HINT_VITA_PVR_INIT, V2H_BOOL_INVERTED },
+    { "VITA_RESOLUTION", SDL_HINT_VITA_RESOLUTION, V2H_PASSTHROUGH },
+#endif
+};
+
+/* TODO: Handle variables that were set previously and later cleared */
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -5624,6 +5624,60 @@ SDL_RemoveTimer(SDL2_TimerID id)
...
+#endif
+#ifdef SDL_PLATFORM_LINUX
+    { "SDL_INPUT_LINUX_KEEP_KBD", SDL_HINT_MUTE_CONSOLE_KEYBOARD, V2H_BOOL_INVERTED },
+#endif
+    { "VITA_DISABLE_TOUCH_BACK", SDL_HINT_VITA_ENABLE_BACK_TOUCH, V2H_BOOL_INVERTED },
+    { "VITA_DISABLE_TOUCH_FRONT", SDL_HINT_VITA_ENABLE_FRONT_TOUCH, V2H_BOOL_INVERTED },
+    { "VITA_MODULE_PATH", SDL_HINT_VITA_MODULE_PATH, V2H_PASSTHROUGH },
+    { "VITA_PVR_OGL", SDL_HINT_VITA_PVR_OPENGL, V2H_BOOL },
+    { "VITA_PVR_SKIP_INIT", SDL_HINT_VITA_PVR_INIT, V2H_BOOL_INVERTED },
+    { "VITA_RESOLUTION", SDL_HINT_VITA_RESOLUTION, V2H_PASSTHROUGH },
```
sezero, line 5653: no vita ifdefs for these?

cgutman: I can add them, but I didn't put them there because the SDL3 hints do nothing on non-Vita platforms anyway.

Though now that you bring it up, I'll ifdef them if only just to save some time comparing against them on non-Vita platforms.

sezero (approved):

src/sdl2_compat.c
```diff
@@ -5624,6 +5624,62 @@ SDL_RemoveTimer(SDL2_TimerID id)
...
+/* TODO: Handle variables that were set previously and later cleared */
+static void SetHintForEnvironmentVariable(const char *name, const char *value)
+{
+    unsigned int i;
+
+    for (i = 0; i < SDL_arraysize(envvars_to_hints); ++i) {
+        if (SDL3_strcmp(name, envvars_to_hints[i].old_envvar) == 0) {
+            switch (envvars_to_hints[i].type) {
+            case V2H_BOOL:
+                value = value ? "1" : "0";
```
sezero, line 5667: Is that supposed to be `*value ? ` instead of `value ? ` ? (Same for the one below)

cgutman: At an initial glance the answer is no, because SDL2 does `SDL_getenv(variable) == NULL` checks rather than actually looking at the contents, so doing the same here is correct.

However, things get more interesting when we look at SDL2's implementation of `SDL_getenv()`. On Windows, the `SDL_getenv()` implementation [treats empty as unset](https://github.com/libsdl-org/SDL/blob/5330041ea9767fc9488e3c46434fb0791002a550/src/stdlib/SDL_getenv.c#L197-L199), but the `getenv()` implementation would return a non-null pointer in that case.

So when discussing SDL2 behavior for a scenario like `FOO= ./sdl2app`, you would actually get different results on Windows vs Linux. Fortunately, I don't think this scenario actually matters in practice since none of these variables are applicable to Windows, so treating these boolean-style hints the Unix way as done here (unset distinct from empty) is fine.

sezero: Can `sep + 1` passed here from `SynchronizeEnvironmentVariables` ever be NULL, though?

cgutman: Not with the current code, because we don't yet support unsetting environment variables through this codepath. I left a todo for that case:
```
/* TODO: Handle variables that were set previously and later cleared */
```

I could check `*value` but that would diverge from SDL2 behavior if someone was doing something like `SDL_HIDAPI_JOYSTICK_DISABLE_UDEV= ./sdl2app`. Under SDL2 and the current state of this PR, this would result in udev being disabled. However, if I change to `*value`, udev would then be _enabled_ in that situation. I _really_ hope nobody is relying on this very subtle difference, but we might as well match SDL2 bug-for-bug here.

slouken:
How does this interact with `renamed_hints`?

cgutman:
> How does this interact with `renamed_hints`?

It shouldn't interact at all (the hints in `renamed_hints` are different than the environment variables in `envvars_to_hints`). Did you have some specific issue in mind?

I did see an issue though with the handling of `renamed_hints` in `SDL2Compat_ApplyQuirks()` while taking a look, so I'll put up a fix for that.

cgutman:
I added similar synchronization logic for `renamed_hints` into `SynchronizeEnvironmentVariables()`. We try to do this in `SDL2Compat_ApplyQuirks()`, but we can't safely call `SDL2_to_SDL3_hint_value()` there because it uses a whole bunch of SDL3 functions. It doesn't matter for most hints, but we won't be able to do the fixup for `SDL_LOGGING` and `SDL_WINDOWS_NO_CLOSE_ON_ALT_F4` which means those values will be set incorrectly until `SynchronizeEnvironmentVariables()` runs for the first time.

I also addressed a related issue where `SDL_VideoInit()` and `SDL_AudioInit()` drivers could be superseded by `SDL_AUDIODRIVER`/`SDL_VIDEODRIVER` environment variables (which isn't how SDL2 behaves).

slouken:
> > How does this interact with `renamed_hints`?
> 
> It shouldn't interact at all (the hints in `renamed_hints` are different than the environment variables in `envvars_to_hints`). Did you have some specific issue in mind?

Well, environment variables act as overrides for hints... it seems like renamed hints should also be treated as renamed environment variables, and I believe the existing code actually does that. So maybe your changes and this code should be combined?

src/sdl2_compat.c
sezero: OK then

cgutman:
> Well, environment variables act as overrides for hints... it seems like renamed hints should also be treated as renamed environment variables, and I believe the existing code actually does that. So maybe your changes and this code should be combined?

I was trying to keep them in separate lists to preserve the (admittedly small) differences between environment variables and hints like:
- Hints (that aren't passed as environment variables) wouldn't show up in the environment if queried via `SDL_getenv()`
- Environment variables would not be reset by `SDL_ResetHints()` but hints would
- Setting a _non-hint_ environment variable via `SDL_SetHint()` wouldn't have any effect (hopefully nobody is relying on this anyway)

With my latest push, I'm following the rules:
- If it comes into sdl2-compat as an environment variable, it gets propagated to SDL3 as an environment variable
- If it comes into sdl2-compat as a hint (via `SDL_SetHint()` and friends), it gets propagated to SDL3 as a hint (this was already happening today)

src/sdl2_compat.c
```diff
@@ -5624,6 +5624,79 @@ SDL_RemoveTimer(SDL2_TimerID id)
...
+    SDL3_SetEnvironmentVariable(env, name, value, true);
+
+    /* Handle environment variables that became hints in SDL3  */
+    for (i = 0; i < SDL_arraysize(envvars_to_hints); ++i) {
+        if (SDL3_strcmp(name, envvars_to_hints[i].old_envvar) == 0) {
+            const char *value3;
+
+            switch (envvars_to_hints[i].type) {
+            case V2H_BOOL:
+                value3 = value ? "1" : "0";
```
slouken, line 5673: There may be exceptions (bugs) but SDL has always generally handled VAR unset and VAR="" to also mean unset.

This code and the code below should probably be something like:
`value3 = SDL3_GetStringBoolean(value, ???) ? "1" : "0";`

cgutman: There was an earlier discussion on this that got hidden when I refactored the code: https://github.com/libsdl-org/sdl2-compat/pull/361#discussion_r1957197273

The `SDL_getenv()` usage in SDL2 for these boolean-style variables wasn't treating unset as equivalent to empty, because it was comparing the pointer itself to null rather than checking that it was non-null and non-empty.

Here are some tests with real SDL2:
```
$ SDL_LOGGING=debug ./test/testgamecontroller
[snip]
DEBUG: Using udev for HIDAPI joystick device discovery
```

```
$ SDL_LOGGING=debug SDL_HIDAPI_JOYSTICK_DISABLE_UDEV= ./test/testgamecontroller
[snip]
DEBUG: udev disabled by SDL_HIDAPI_JOYSTICK_DISABLE_UDEV
```

So using `SDL3_GetStringBoolean(value)` or `*value` wouldn't match SDL2 behavior. I really hope nobody is (intentionally) relying on this behavior, but I think we're trying to stay bug-compatible with SDL2 as much as possible.

slouken: Fair enough. FYI, SDL3 generally treats variables equal to "" as unset.

PR #362, opened 2025-02-15
init: Print the required SDL3 version even if a function is unavailable

cgutman:
This PR moves the loading of most SDL3 symbols after the point where we perform our required version check. In the event that use of a new SDL3 function is added and required by sdl2-compat (like we did in 2.30.54), we want to be able to print the friendly SDL3 required version text rather than just barfing with the name of a missing SDL3 function.

As a side-effect, this strictly enforces the rule of not calling into SDL3 functions (except `SDL_GetVersion()`) during `LoadSDL3()` and exposed one place where we were doing that.

Before:
```
SDL_StretchSurface missing in SDL3 library.
Aborted (core dumped)
```

After:
```
sdl2-compat 2.32.51: SDL3 library is too old (have 3.1.7, but need at least 3.2.4).
Aborted (core dumped)
```

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -282,6 +282,19 @@ static bool SDL2Compat_strequal(const char *a, const char *b)
     return true;
 }
 
+/* you can use SDL3_strrchr once we're past startup. */
+static char *SDL2Compat_strrchr(const char *string, int c)
+{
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #363, opened 2025-02-16
audio: Fix invalid device index in SDL_AUDIODEVICEADDED

cgutman:
`testaudiohotplug` is broken because the `SDL_AUDIODEVICEADDED` event is not getting fixed up to convert the `SDL_AudioDeviceID` to an index. This causes a bunch of errors during device enumeration.

```
INFO: Available audio drivers:
INFO: 0: pipewire
INFO: 1: pulseaudio
INFO: 2: alsa
INFO: 3: jack
INFO: 4: disk
INFO: 5: dummy
INFO: Select a driver with the SDL_AUDIODRIVER environment variable.
INFO: Using audio driver: pipewire
ERROR: Got new output device at index 15, but failed to get the name: Parameter 'index' is invalid
ERROR: Got new capture device at index 18, but failed to get the name: Parameter 'index' is invalid
ERROR: Got new output device at index 23, but failed to get the name: Parameter 'index' is invalid
ERROR: Got new capture device at index 26, but failed to get the name: Parameter 'index' is invalid
```

With this change, it works properly again:
```
INFO: Available audio drivers:
INFO: 0: pipewire
INFO: 1: pulseaudio
INFO: 2: alsa
INFO: 3: jack
INFO: 4: disk
INFO: 5: dummy
INFO: Select a driver with the SDL_AUDIODRIVER environment variable.
INFO: Using audio driver: pipewire
INFO: New output audio device at index 0: Built-in Audio Analog Stereo
INFO: Opened 'Built-in Audio Analog Stereo' as 2
INFO: New capture audio device at index 0: Built-in Audio Analog Stereo
```

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1572,6 +1572,20 @@ static SDL_Keycode SDL3KeycodeToSDL2Keycode(SDL_Scancode scancode, SDL_Keycode k
     return SDL_SCANCODE_TO_KEYCODE(scancode);
 }
 
+static int GetIndexFromAudioDeviceInstance(SDL_AudioDeviceID devid, bool recording)
+{
+    AudioDeviceList *list = recording ? &AudioSDL3RecordingDevices : &AudioSDL3PlaybackDevices;
+    int i;
+
+    for (i = 0; i < list->num_devices; i++) {
+        if (devid == list->devices[i].devid) {
+            return i;
+        }
+    }
+
+    return -1;
+}
+
 /* (current) strategy for SDL_Events:
    in sdl12-compat, we built our own event queue, so when the SDL2 queue is pumped, we
    took the events we cared about and added them to the sdl12-compat queue, and otherwise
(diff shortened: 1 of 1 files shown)


slouken:
Nice fix, thanks!

src/sdl2_compat.c
```diff
@@ -1833,6 +1850,23 @@ Event2to3(const SDL2_Event *event2, SDL_Event *event3)
     case SDL_EVENT_JOYSTICK_BATTERY_UPDATED:
         /* This should never happen, but see Event3to2() for details */
         break;
+    case SDL_EVENT_AUDIO_DEVICE_ADDED:
+        if (event2->adevice.iscapture) {
+            if (event2->adevice.which >= 0 &&
```
sezero, line 1855: this `which>=0`check is bogus: which member is an Uint32. (same with the one below.) Should we remove?

slouken: Yes please.

sezero: Done

PR #369, opened 2025-02-17
env: Handle syncing variables that were unset and renamed hints/variables through SDL_setenv()

cgutman:
This covers hopefully the last remaining things for hint/variable handling:
- Clearing hints/variables that were previously set after an app does `unsetenv()` later on (fixing the `TODO` from the previous PR)
- Handling renamed hints/variables that are set through `SDL_setenv()`


slouken:
Merged, thanks!

PR #377, opened 2025-02-20
Small performance optimizations

cgutman:
This PR has a few minor performance optimizations to reduce calls to `SDL3_malloc()` and avoid unnecessary renderer flushes in `SDL_RenderDrawRects()`.


slouken:
Merged, thanks!

PR #378, opened 2025-02-21
mouse: Accumulate fractional mouse data

### LINKED ISSUE #372: SDL_MOUSEMOTION events report incorrect relative motion
I'm on Arch Linux using sdl2-compat 2.30.52-1, X11. Tested that this works correctly on sdl2 2.16.0.

When moving the cursor single pixels at a time, the xrel and yrel fields report 0 instead of 1. Other cases are incorrect as well, there seems to be some accumulation going on that isn't reset/rounded properly. If you move the cursor large amounts and then by single pixels, xrel/yrel will still report those large coordinates in future mouse motion events.

cgutman:
This PR accumulates the floating point values returned in `SDL_EVENT_MOUSE_MOTION`, `SDL_EVENT_MOUSE_WHEEL`, and `SDL_GetRelativeMouseState()`.

For the events, I'm using global variables rather than per-window state which I _think_ should fine because only one window is receiving mouse input at a time.

Fixes #372

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1028,6 +1028,12 @@ static SDL_mutex *EventWatchListMutex = NULL;
 static SDL2_LogOutputFunction LogOutputFunction2 = NULL;
 static EventFilterWrapperData *EventWatchers2 = NULL;
 static SDL2_bool relative_mouse_mode = SDL2_FALSE;
+static float relative_mouse_state_x_frac = 0.0f;
+static float relative_mouse_state_y_frac = 0.0f;
+static float relative_mouse_event_x_frac = 0.0f;
+static float relative_mouse_event_y_frac = 0.0f;
+static float mouse_wheel_event_x_frac = 0.0f;
+static float mouse_wheel_event_y_frac = 0.0f;
 static SDL_JoystickID *joystick_instance_list = NULL;
 static int num_joystick_instances = 0;
 static SDL_JoystickID *joystick_list = NULL;
@@ -1599,6 +1605,22 @@ static int GetIndexFromAudioDeviceInstance(SDL_AudioDeviceID devid, bool recordi
     return -1;
 }
 
+static int AccumulateFloatValueToInteger(float *frac, float value)
+{
+    float intval;
+
+    /* Reset the accumulated fractional value if the sign changes */
+    if ((*frac < 0.0f && value > 0.0f) || (*frac > 0.0f && value < 0.0f)) {
+        *frac = 0.0f;
+    }
+
+    /* Accumulate the fractional portion that is truncated by integer conversion */
+    *frac += value;
+    *frac = SDL3_modff(*frac, &intval);
+
+    return (int)intval;
+}
+
 /* (current) strategy for SDL_Events:
    in sdl12-compat, we built our own event queue, so when the SDL2 queue is pumped, we
    took the events we cared about and added them to the sdl12-compat queue, and otherwise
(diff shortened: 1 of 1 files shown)


slouken:
Is SDL_modff() helpful for this?

slouken:
https://gist.github.com/slouken/7c41d29a252d58778794311f555ebe8c

cgutman:
> Is SDL_modff() helpful for this?

We could use it, but it doesn't seem like it would end up much different in complexity than the current solution.

slouken:
It wouldn't be dependent on the current rounding mode when you're casting to int...

slouken:
Also, it would be nifty to actually use the fractional accumulation gist I wrote for GDC. :)

cgutman:
> It wouldn't be dependent on the current rounding mode when you're casting to int...

Good point, updated the PR to use `SDL_modff()`.

src/sdl2_compat.c
```diff
@@ -1685,14 +1714,14 @@ static SDL2_Event *Event3to2(const SDL_Event *event3, SDL2_Event *event2)
...
-            motion->yrel = (Sint32)event3->motion.yrel;
+            motion->xrel = (Sint32)AccumulateFloatValueToInteger(&relative_mouse_event_x_frac, event3->motion.xrel);
+            motion->yrel = (Sint32)AccumulateFloatValueToInteger(&relative_mouse_event_y_frac, event3->motion.yrel);
         } else {
             SDL2_MouseMotionEvent *motion = &event2->motion;
             motion->x = (Sint32)event3->motion.x;
             motion->y = (Sint32)event3->motion.y;
-            motion->xrel = (Sint32)event3->motion.xrel;
-            motion->yrel = (Sint32)event3->motion.yrel;
+            motion->xrel = (Sint32)AccumulateFloatValueToInteger(&relative_mouse_event_x_frac, event3->motion.xrel);
```
slouken, line 1723: Cast may not be needed here, definitely not needed below casting to int.

src/sdl2_compat.c
```diff
@@ -1685,14 +1714,14 @@ static SDL2_Event *Event3to2(const SDL_Event *event3, SDL2_Event *event2)
...
+            motion->yrel = (Sint32)AccumulateFloatValueToInteger(&relative_mouse_event_y_frac, event3->motion.yrel);
         }
         break;
```
slouken, line 1726: Maybe drop the event if xrel and yrel are zero?

src/sdl2_compat.c
cgutman: I'll ditch the casts and see if CI stays happy on all platforms.

src/sdl2_compat.c
cgutman: I think we'd need more extensive refactoring to support `Event3to2()` deciding that an event should be dropped  entirely. Functions like `SDL_WaitEventTimeout()` and `SDL_PeepEvents()` are calling it and assuming a 1:1 mapping between SDL2 and SDL3 events.

If we woke up from `SDL_WaitEventTimeout()` with an event that we later decided to drop, we'd need to go back to sleep again while remembering how much time we had slept for previously.

There's also other corner cases like emulated mouse input via touch where we [purposely suppress the relative motion](https://github.com/libsdl-org/SDL/blob/b5297de56fc20af1f5c6d0decb1da7f3ade369e6/src/events/SDL_mouse.c#L743-L747) to avoid having massive relative deltas when the user taps different places on screen.

I think it's probably best to just send the extra events unless we discover something breaking due to them. That's also how it worked for `SDL_MOUSEWHEEL` back when we added the `preciseX` and `preciseY` fields. You'd get `SDL_MOUSEWHEEL` events with `x=0, y=0` (and non-zero `preciseX` and/or `preciseY`) until enough accumulated to roll x/y over to the next integer value.

slouken: Sounds good.

slouken:
Merged, thanks!

PR #380, opened 2025-02-23
audiocvt: Fix apps that look at filters[0] to see if conversion is needed

### LINKED ISSUE #379: sound issues with sauerbraten
Sound scratches in `sauerbraten-linux`
Not happening with legacy `sdl2`

cgutman:
We don't have old-style SDL2 filters to put in this array anymore, but we'll stuff a decoy in there to fix apps that are poking at internal things that they shouldn't be.

Fixes #379


src/sdl2_compat.c
```diff
@@ -10487,10 +10487,16 @@ SDL_BuildAudioCVT(SDL_AudioCVT *cvt,
             &ap,
             sizeof(ap));
 
-        cvt->filters[0] = NULL;
+        /* Saurbraten pokes around in undocumented SDL2 internals to see if format conversion
+         * is needed instead of just reading the documented field. Fake a filter in the first
+         * position if we're going to need audio conversion, so apps that do this will work. */
+        cvt->filters[0] = (SDL_AudioFilter)1;
+        cvt->filter_index = 1;
```
cgutman, line 10494: Saurbraten doesn't look at `filter_index`, but might as well set it if we going to play games with `filters` too...

icculus:
Just in case an app is going to go completely cowboy, maybe this should provide an actual function that just runs the conversion; we don't ever call it ourselves but it'll do the right thing if called...?

cgutman:
> Just in case an app is going to go completely cowboy, maybe this should provide an actual function that just runs the conversion; we don't ever call it ourselves but it'll do the right thing if called...?

Okay, I may have gone overboard with compat handling, but it should now work as SDL2 did even if they're messing with `src_format` after `SDL_BuildAudioCVT()`, invoking filters directly, or even chaining their own filters (as long as they don't collide with the `AudioParam` we stash there).

cgutman:
Actually maybe it's just simpler to implement `SDL_ConvertAudio()` the way SDL2 did where it invokes the filter chain and move the conversion logic into a real filter. Let me try that.

cgutman:
The `arm32` build failure looks like a race condition. It will probably pass if someone reruns it.

PR #383, opened 2025-02-24
events: Fix text input being re-enabled for each new window created

cgutman:
The SDL2 docs were incorrect about this (fixed in https://github.com/libsdl-org/SDL/commit/92cadaf917b5c77d668a09aae782234b9d0c6efe). The actual behavior was that text input was enabled by default during video init, not each time a window was created.

Apps that call `SDL_StopTextInput()` [before window creation](https://github.com/daid/SeriousProton/blob/d453aa6eb124a68d90b0246fe00b93f9c02df671/src/engine.cpp#L120-L122) were unexpectedly having it re-enabled behind their backs.

Fixes broken key bindings in EmptyEpsilon.


slouken:
Merged, thanks!

PR #386, opened 2025-02-24
Drop some SDL3 events that incorrectly overlap SDL2 events

Kontrabant:
The SDL3 keyboard added event number overlaps the SDL2 SDL_TEXTEDITING_EXT event, which can cause a client to crash if it tries processing the keyboard added event as an IME text candidate event with no validation.

Just drop the keyboard added/removed events from the queue, as they have no SDL2 equivalent and would only be ignored.

Caught this with the GOG version of VVVVVV, which was crashing on a keyboard added event due to trying to process it as an IME text event and calling strlen with a null string parameter.


slouken:
Whoops, good catch, thanks!

PR #387, opened 2025-02-26
audio: Fix available audio stream byte count

cgutman:
Fixes the issue with `SDL_GetQueuedAudioSize()` behavior mentioned here: https://github.com/libsdl-org/sdl2-compat/issues/329#issuecomment-2675315002
> SDL_GetQueuedAudioSize() reports bytes in the queue...in SDL2, this is converted bytes, but in SDL3 this is data that has not yet been converted...


src/sdl2_compat.c
```diff
@@ -7089,7 +7089,21 @@ SDL_AudioStreamClear(SDL2_AudioStream *stream2)
...
+    }
+
+    if (!SDL3_GetAudioStreamFormat(stream2->stream3, &src_spec, &dst_spec)) {
+        return 0;
+    }
+
+    src_size = src_spec.channels * src_spec.freq * SDL_AUDIO_BYTESIZE(src_spec.format);
+    dst_size = dst_spec.channels * dst_spec.freq * SDL_AUDIO_BYTESIZE(dst_spec.format);
+
+    return (Uint64)SDL3_GetAudioStreamAvailable(stream2->stream3) * src_size / dst_size;
```
slouken, line 7106: Do you need to cast back to int for the final value?

slouken:
Merged, thanks!

PR #388, opened 2025-02-26
sensors: Implement GetSensorDataWithTimestamp functions

cgutman:
These are implemented by snooping the sensor events in our event filter, just as the FIXME comment suggested.

These functions didn't get much uptake but [Vita3k](https://github.com/Vita3K/Vita3K/blob/3d53d7cb93ada60076925d73aa20d39c08ea4409/vita3k/motion/src/motion.cpp#L115) and [node-sdl](https://github.com/kmamal/node-sdl/blob/8a2c7628e31c320bb718881f8139900cce15b1eb/src/native/sensor.cpp#L90) are known users of them.


slouken:
Merged, thanks!

PR #389, opened 2025-02-26
audio: Fix device ID in SDL_AUDIODEVICEREMOVED

cgutman:
`SDL_AudioDeviceID` changed semantics from just identifying an open audio stream in SDL2 to also identifying physical audio devices in SDL3. We need to switch it back into an SDL2-style `SDL_AudioDeviceID` before handing it over to the SDL2 app or it may try to close a non-existent audio stream (or worse).


slouken:
Merged, thanks!

PR #390, opened 2025-02-26
audio: Fix incorrect value returned from SDL_AudioStreamAvailable()

cgutman:
#387 put adjustment logic for input vs output size changes for `SDL_GetQueuedAudioSize()` in `SDL_AudioStreamAvailable()`, but this isn't correct for callers that are using the audio stream API directly. The real issue was that `SDL_GetQueuedAudioSize()` shouldn't have been directly returning the value from `SDL_AudioStreamAvailable()`, so move the size adjustment logic into `SDL_GetQueuedAudioSize()` instead (while preserving the fix from #387 for a `NULL` stream passed to `SDL_AudioStreamAvailable()`).


slouken:
Merged, thanks!

PR #391, opened 2025-02-27
audio: Close devices when the audio subsystem quits

### LINKED ISSUE #276: Kdenlive does not play sound when USB audio interface is connected
Arch Linux, system fully up to date at the time of this writing.
Since a couple of days ago, roughly aligned with the day when Arch pushed SDL3 update, Kdenlive no longer plays audio WHILE my USB sound card (Audient EVO4) is connected. When starting up it complains with the following message:

MLT: "[consumer sdl2_audio ] SDL failed to open audio"

This only happens with Kdenlive. All other programs that I have installed play sound normally.
When I unplug my Audient USB interface and start KDenlive - sound plays fine and this error is not reported.
Further: When I have Kdenlive running with sound like that I can plug in my Audient at that point and the sound will switch to it (I have headphones on it) and it will play fine for a couple of minutes and then it's gone again.

I will provide more details if you tell me what to search for.

cgutman:
Kdenlive relies on this behavior to avoid reaching the open audio device limit.

Fixes #276

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -6147,6 +6147,37 @@ static int InitSubsystemInternal(SDL_InitFlags flags)
     return result;
 }
 
+static void PostQuitSubsystem(SDL_InitFlags quit_flags)
+{
+    /* Mask out the subsystems that didn't actually quit */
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #392, opened 2025-02-28
Improve accuracy of audio/video subsystem init/quit functions

cgutman:
These weren't quite working like they did in SDL2, particularly around switching drivers by reinitializing and quitting when additional references were present.


slouken:
Merged, thanks!

Issue #394, opened 2025-03-02
ioq3 not scaled up with wayland output


jpalus:
I can't really tell if the issue is in sdl2-compat or SDL3 but I'll start here.

Running [ioq3](https://github.com/ioquake/ioq3) within wayland compositor (sway) on Linux, native display resolution 2560x1440 and in-game mode 1920x1080 gives following results with "original" SDL2 library:
- video driver `x11`: game is not scaled up and is anchored at lower left corner
- video driver `wayland`: game is scaled up to native resolution

Now with SDL3 3.2.6 and sdl2-compat 2.32.52 both drivers give the same result -- game not scaled up anchored at lower left corner same as with `x11` above.

Not sure if it's of any use but that's what I found regarding window/mode initialization in ioq3:

window is created with SDL_WINDOW_FULLSCREEN flag set:
https://github.com/ioquake/ioq3/blob/3fb9006e6461b877b67781cd0f0d94032804a475/code/sdl/sdl_glimp.c#L691-L692

then mode is set:
https://github.com/ioquake/ioq3/blob/3fb9006e6461b877b67781cd0f0d94032804a475/code/sdl/sdl_glimp.c#L698-L714

slouken:
I think this is a game bug. It sets the mode but then doesn't check to see if the mode is actually supported or whether it got the size it expected.

That said, I am seeing issues with creating windows fullscreen in sdl2-compat, so I'll fix those.

Kontrabant:
The sdl2-compat mode switching code will only try to switch the mode if an exclusive fullscreen mode was previously set, which doesn't happen in SDL3 when creating a window with the fullscreen flag, since it's desktop fullscreen.

If the check is removed in ApplyFullscreenMode, ioq3 works correctly.

slouken:
The x11 bug is caused by the game assuming that it got the mode it requested, which doesn't happen (and isn't possible) under XWayland.

slouken:
> The x11 bug is caused by the game assuming that it got the mode it requested, which doesn't happen (and isn't possible) under XWayland.

Actually, @Kontrabant, looking at the X11 SDL code it looks like mode switches under XWayland are intended to work, but what I'm seeing is that the window remains the full desktop size. Is that something we broke, or is that a regression or change to XWayland?

slouken:
[...]

PR #395, opened 2025-03-02
Fixed creating exclusive fullscreen windows

slouken:
Also fixed ApplyFullscreenMode(), which should apply the closest mode to the requested mode, or if no mode was requested, then the closest to the window size.

Also fixed the output of SDL_GetWindowDisplayMode(), which should return the mode that will actually be used.

Fixes https://github.com/libsdl-org/sdl2-compat/issues/394

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -7698,49 +7698,35 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
 
 #define PROP_WINDOW_FULLSCREEN_MODE "sdl2-compat.window.fullscreen-mode"
 
-static int
-ApplyFullscreenMode(SDL_Window *window)
+static int ApplyFullscreenMode(SDL_Window *window)
 {
-    SDL2_DisplayMode *property = (SDL2_DisplayMode *) SDL3_GetPointerProperty(SDL3_GetWindowProperties(window), PROP_WINDOW_FULLSCREEN_MODE, NULL);
+    /* Always try to enter fullscreen on the current display */
+    const SDL_DisplayID displayID = SDL3_GetDisplayForWindow(window);
+    SDL2_DisplayMode *property = (SDL2_DisplayMode *)SDL3_GetPointerProperty(SDL3_GetWindowProperties(window), PROP_WINDOW_FULLSCREEN_MODE, NULL);
     SDL_DisplayMode mode;
+
     SDL3_zero(mode);
     if (property) {
-        /* Always try to enter fullscreen on the current display */
-        const SDL_DisplayID displayID = SDL3_GetDisplayForWindow(window);
-
-        /* The SDL2 refresh rate is rounded off, and SDL3 checks that the mode parameters match exactly, so try to find the closest matching SDL3 mode. */
-        SDL3_GetClosestFullscreenDisplayMode(displayID, property->w, property->h, (float)property->refresh_rate, false, &mode);
+        mode.w = property->w;
+        mode.h = property->h;
+        mode.refresh_rate = (float)property->refresh_rate;
+    } else {
+        SDL3_GetWindowSize(window, &mode.w, &mode.h);
     }
+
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -7698,32 +7698,36 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
...
-        /* Always try to enter fullscreen on the current display */
-        const SDL_DisplayID displayID = SDL3_GetDisplayForWindow(window);
-
-        /* The SDL2 refresh rate is rounded off, and SDL3 checks that the mode parameters match exactly, so try to find the closest matching SDL3 mode. */
-        SDL3_GetClosestFullscreenDisplayMode(displayID, property->w, property->h, (float)property->refresh_rate, false, &mode);
+        mode.w = property->w;
+        mode.h = property->h;
+        mode.refresh_rate = (float)property->refresh_rate;
+    } else {
+        const SDL_DisplayMode *desktop_mode = SDL3_GetDesktopDisplayMode(displayID);
```
cgutman, line 7714: I don't think this part is correct. SDL2 uses similar logic to our `SDL_GetWindowDisplayMode()` to determine the fullscreen resolution, but it will fail outright if a suitable resolution can't be found instead of picking desktop mode.

cgutman: Ah, never mind. I saw that you are overriding the dimensions with the window size. Though, I think calling `SDL3_GetDesktopDisplayMode()` isn't necessary  because `SDL3_GetClosestFullscreenDisplayMode()` will use the desktop refresh rate already if you pass 0.0f.

src/sdl2_compat.c
```diff
@@ -7698,32 +7698,36 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
...
-        SDL_DisplayID displayID;
         int ret;
 
-        displayID = SDL3_GetDisplayForWindow(window);
-        if (!displayID) {
-            displayID = SDL3_GetPrimaryDisplay();
-        }
-
         /* FIXME: at least set a valid fullscreen mode */
         list = SDL3_GetFullscreenDisplayModes(displayID, &count);
```
cgutman, line 7732: I think all this fallback logic here should be removed though since SDL2 just fails if a compatible mode cannot be found/set.

src/sdl2_compat.c
slouken: Good call, I'll remove that.

src/sdl2_compat.c
slouken: Done!

src/sdl2_compat.c
```diff
@@ -8211,12 +8209,15 @@ SDL_CreateWindow(const char *title, int x, int y, int w, int h, Uint32 flags)
...
+    bool exclusive_fullscreen = false;
 
     CheckEventFilter();
 
-    if (flags & SDL2_WINDOW_FULLSCREEN_DESKTOP) {
+    if ((flags & SDL2_WINDOW_FULLSCREEN_DESKTOP) == SDL2_WINDOW_FULLSCREEN_DESKTOP) {
         flags &= ~SDL2_WINDOW_FULLSCREEN_DESKTOP;
         flags |= SDL_WINDOW_FULLSCREEN; /* This is fullscreen desktop for new windows */
+    } else if (flags & SDL_WINDOW_FULLSCREEN) {
+        exclusive_fullscreen = true;
```
cgutman, line 8220: I think we also need to unset the `SDL_WINDOW_FULLSCREEN` here to get the correct mode selected. When I was testing this commit creating a 640x480 fullscreen window, the resulting fullscreen mode ended up being the desktop mode. When SDL3 enters fullscreen desktop mode during window creation, the call to `SDL3_GetWindowSize()` will return the desktop dimensions and cause us to switch to the wrong mode.

slouken: Oh, good point. We should probably also mark the window hidden if we’re about to set a full screen mode.

PR #397, opened 2025-03-03
events: Add support for printing SDL2 events

cgutman:
Setting `SDL_HINT_EVENT_LOGGING` only logs from the SDL3 side, which isn't always a 1:1 mapping with what the SDL2 app is going to see. This PR forklifts the `SDL_LogEvent()` function from SDL2 with some tweaks and brings it into sdl2-compat to allow us to see exactly what SDL2 events are making it to the app for easier debugging.

Right now I've just wired it to the `SDL2_EVENT_LOGGING` hint, leaving `SDL_EVENT_LOGGING` to activate the regular SDL3 logging.


slouken:
Looks good, thanks!

PR #403, opened 2025-03-09
Use SDL3's integer mouse mode hint

### LINKED ISSUE #372: SDL_MOUSEMOTION events report incorrect relative motion

cgutman:
This reverts https://github.com/libsdl-org/sdl2-compat/pull/378 and uses the new SDL3 integer mouse mode hint from https://github.com/libsdl-org/SDL/pull/12497 (further modified in https://github.com/libsdl-org/SDL/pull/12551) instead.

I also searched GitHub for popular SDL2 apps that support fractional wheel events and preemptively quirked them so they will continue to receive them. Unquirked apps will just receive wheel events when they accumulate to an integer value (preserving compatibility but losing smooth scrolling). We can quirk others if users report loss of smooth scrolling with sdl2-compat.

Fixes #372

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1309,6 +1309,10 @@ SDL2Compat_InitOnStartupInternal(void)
     SDL3_SetHint("SDL_VIDEO_SYNC_WINDOW_OPERATIONS", "1");
     SDL3_SetHint("SDL_VIDEO_X11_EXTERNAL_WINDOW_INPUT", "0");
 
+    /* Emulate both integer mouse coordinates and integer mouse wheel deltas for maximum compatibility.
+       Apps that use preciseX/Y for smooth scrolling can be quirked to get fractional wheel deltas. */
+    SDL3_SetHint("SDL_MOUSE_INTEGER_MODE", "3");
+
     // Pretend Wayland doesn't have fractional scaling by default.
     // This is more compatible with applications that have only been tested under X11 without high DPI support.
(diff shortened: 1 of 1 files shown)


slouken:
We should wait to merge this until we get closer to the next SDL3 release.

sezero:
This will need to require 3.2.10 at runtime (when it's released)

slouken:
> This will need to require 3.2.10 at runtime (when it's released)

Yep, I'll take care of this when I merge it.

cgutman:
Thanks, updated the PR for that change.

v1993:
OpenTTD relies on both fractional values (for smooth two-finger viewport dragging) and accumulated integer values (for scrolling in menus). Should it be quirked?

slouken:
> OpenTTD relies on both fractional values (for smooth two-finger viewport dragging) and accumulated integer values (for scrolling in menus). Should it be quirked?

Yes, feel free to submit a tested PR.

v1993:
@slouken I've tried it and can't get both accumulated integer and precise fractional scrolling to work at the same time (default is jumpy 2-finger dragging, `SDL_MOUSE_INTEGER_MODE=1` breaks menu scrolling on touchpad). Is it possible to accurately emulate old behavior where accumulation was done for integer but not precise values?

PR #404, opened 2025-03-09
Fixed crash during controller hotplug in RetroArch

slouken:
RetroArch calls SDL_PumpEvents() followed by SDL_PeepEvents(), so we need to update the joystick list inside of Event3to2(). This event doesn't happen often, so this should be fine.


slouken:
Added @smcv for visibility

PR #411, opened 2025-03-14
Pass original name through hints callbacks

madebr:
Pass the original hint name to the hint callbacks, instead of renamed ones.

Please review this for oversights.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -224,6 +224,13 @@ do { \
 static bool WantDebugLogging = false;
 static SDL_InitState InitSDL2CompatGlobals;
 
+typedef struct HintCallbackInfo
+{
+    char *name;
+    SDL_HintCallback callback;
+    void *userdata;
+} HintCallbackInfo;
+static SDL_PropertiesID hint_callbacks = 0;
 
 static char *
 SDL2COMPAT_stpcpy(char *dst, const char *src)
@@ -621,6 +628,7 @@ static const char *SDL2_to_SDL3_hint(const char *name)
     return name;
 }
 
+/* This function must be kept in sync with SDL3_to_SDL2_hint_value */
 static const char *SDL2_to_SDL3_hint_value(const char *name, const char *value, bool *free_value)
 {
     *free_value = false;
@@ -641,6 +649,34 @@ static const char *SDL2_to_SDL3_hint_value(const char *name, const char *value, 
             return value3;
         } else if (SDL3_strcmp(name, "SDL_WINDOWS_NO_CLOSE_ON_ALT_F4") == 0) {
             /* Invert the boolean value for SDL3 */
+            return (*value == '0' || SDL3_strcasecmp(value, "false") == 0) ? "1" : "0";
+        }
+    }
+
+    return value;
+}
+
+/* This function must be kept in sync with SDL2_to_SDL3_hint_value */
+static const char *SDL3_to_SDL2_hint_value(const char *name, const char *value, bool *free_value)
+{
+    *free_value = false;
+
+    if (name && value && *value) {
+        if (SDL3_strcmp(name, SDL_HINT_LOGGING) == 0) {
+            /* Rewrite numeric priorities for SDL3 */
+            char *value2 = SDL3_strdup(value);
+            if (value2) {
+                char *sep;
+                for (sep = SDL3_strchr(value2, '='); sep; sep = SDL3_strchr(sep + 1, '=')) {
+                    if (SDL3_isdigit(sep[1]) && sep[1] != '0') {
+                        sep[1] = '0' + SDL3_atoi(&sep[1]) - 1;
+                    }
+                }
+            }
+            *free_value = true;
+            return value2;
+        } else if (SDL3_strcmp(name, "SDL_WINDOWS_NO_CLOSE_ON_ALT_F4") == 0) {
+            /* Invert the boolean value for SDL2 */
             return (*value == '0' || SDL3_strcasecmp(value, "false") == 0) ? "1" : "0";
         }
     }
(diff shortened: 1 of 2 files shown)


src/sdl2_compat.c
```diff
@@ -681,18 +690,80 @@ SDL_GetHintBoolean(const char *name, SDL2_bool default_value)
     return SDL3_GetHintBoolean(SDL2_to_SDL3_hint(name), default_value) ? SDL2_TRUE : SDL2_FALSE;
 }
 
-/* FIXME: callbacks may need tweaking ... */
+static void SDLCALL hint_callback_wrapper(void *userdata, const char *name, const char *oldValue, const char *newValue)
+{
+    HintCallbackInfo *info = (HintCallbackInfo *) userdata;
+    info->callback(info->userdata, info->name, oldValue, newValue);
```
cgutman, line 696: We need to convert `oldValue` and `newValue` with `SDL2_to_SDL3_hint_value()` too.

src/sdl2_compat.c
```diff
@@ -681,18 +690,80 @@ SDL_GetHintBoolean(const char *name, SDL2_bool default_value)
...
+            name = callback_info->name;
+            if (prev_next) {
+                *prev_next = callback_info->next;
+            } else if (callback_info->next) {
+                SDL3_SetPointerProperty(hint_callbacks, name, callback_info->next);
+            } else {
+                /* FIXME: Cannot remove a property */
+                SDL3_SetPointerProperty(hint_callbacks, name, NULL);
+            }
+            SDL_free(callback_info->name);
```
cgutman, line 761: We could use `name+callback+userdata` as the property name and that would avoid having to manage a linked list to support multiple callbacks for the same hint.

src/sdl2_compat.c
```diff
@@ -681,18 +690,80 @@ SDL_GetHintBoolean(const char *name, SDL2_bool default_value)
...
+        HintCallbackInfo *callback_info = GetHintCallbackInfo(name, callback, userdata, &prev_next);
+
+        SDL3_RemoveHintCallback(SDL2_to_SDL3_hint(name), hint_callback_wrapper, callback_info);
+
+        if (callback_info) {
+            name = callback_info->name;
+            if (prev_next) {
+                *prev_next = callback_info->next;
+            } else if (callback_info->next) {
+                SDL3_SetPointerProperty(hint_callbacks, name, callback_info->next);
```
cgutman, line 756: I think we need `SDL_SetPointerPropertyWithCleanup()` to avoid leaking `callback_info->name` and `callback_info` if somebody calls `SDL_Quit()` without unregistering their hint callbacks.

src/sdl2_compat.c
madebr: Can you please re-review my latest iteration?
old and new names are converted, goes all-in on properties and no longer uses linked lists.

src/sdl2_compat.c
```diff
@@ -684,18 +720,100 @@ SDL_GetHintBoolean(const char *name, SDL2_bool default_value)
...
-/* FIXME: callbacks may need tweaking ... */
+static void SDLCALL
+hint_callback_wrapper(void *userdata, const char *name, const char *oldValue, const char *newValue)
+{
+    HintCallbackInfo *callback_info = (HintCallbackInfo *) userdata;
+    const char *sdl2OldValue, *sdl2NewValue;
+    bool freeSdl2OldValue, freeSdl2NewValue;
+
+    sdl2OldValue = SDL3_to_SDL2_hint_value(name, oldValue, &freeSdl2OldValue);
+    sdl2NewValue = SDL3_to_SDL2_hint_value(name, newValue, &freeSdl2NewValue);
```
cgutman, line 731: `SDL3_to_SDL2_hint_value()` expects the SDL2 hint name, but it looks like it's getting the SDL3 hint name here.

src/sdl2_compat.c
```diff
@@ -684,18 +720,100 @@ SDL_GetHintBoolean(const char *name, SDL2_bool default_value)
...
+}
+
+static char *
+CreateHintCallbackPropertyName(const char *name, SDL_HintCallback callback, void *userdata) {
+    size_t len_property_name = SDL_strlen(name) + 1 + 2 * sizeof(callback) + 1 + 2 * sizeof(userdata) + 1;
+    char *property_name = (char *)SDL3_malloc(len_property_name);
+    if (!property_name) {
+        return NULL;
+    }
+    SDL3_snprintf(property_name, len_property_name, "%s_%x_%x", name, (void *)callback, (void *)userdata);
```
cgutman, line 748: We want `%p` instead of `%x` because the width will depend on the bitness of the platform.

madebr: I did use `%x` because we know an upper bound on the required length (`2 * sizeof(void *)`).
When using `%p`, it depends on the platform whether you get a `0x` prefix or none.

src/sdl2_compat.c
madebr: Indeed. sdl2 names are available in `((HintCallbackInfo *)userdata)->name`;

src/sdl2_compat.c
madebr: You're probably right: `%x`will treat the argument as a 32-bit integer where we need more bits.

cgutman (commented):
We should use `SDL3_free()` instead of `SDL_free()`. Right now there's a mix of both in these changes. Other than that, LGTM.

madebr:
> We should use `SDL3_free()` instead of `SDL_free()`. Right now there's a mix of both in these changes. Other than that, LGTM.

Thanks! That was an oversight.
I applied the following diff in my latest iteration:
```diff
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -753,8 +753,8 @@ static void SDLCALL
 HintCallbackInfoPropertyCleanup(void *userdata, void *value)
 {
     HintCallbackInfo *callback_info = (HintCallbackInfo *) value;
-    SDL_free(callback_info->name);
-    SDL_free(callback_info);
+    SDL3_free(callback_info->name);
+    SDL3_free(callback_info);
 }
 
 SDL_DECLSPEC void SDLCALL
@@ -777,10 +777,10 @@ SDL_AddHintCallback(const char *name, SDL_HintCallback callback, void *userdata)
             SDL3_free(property_name);
             return;
         }
-        callback_info->name = SDL_strdup(name);
+        callback_info->name = SDL3_strdup(name);
         if (!callback_info->name) {
-            SDL_free(property_name);
-            SDL_free(callback_info);
+            SDL3_free(property_name);
+            SDL3_free(callback_info);
             return;
         }
         callback_info->callback = callback;
@@ -790,7 +790,7 @@ SDL_AddHintCallback(const char *name, SDL_HintCallback callback, void *userdata)
         callback = hint_callback_wrapper;
         userdata = callback_info;
 
-        SDL_free(property_name);
+        SDL3_free(property_name);
     }
 
     /* this returns an int of 0 or -1 in SDL3, but SDL2 it was void (even if it failed). */
@@ -810,7 +810,7 @@ SDL_DelHintCallback(const char *name, SDL_HintCallback callback, void *userdata)
         callback_info = (HintCallbackInfo *) SDL3_GetPointerProperty(hint_callbacks, property_name, NULL);
         SDL3_RemoveHintCallback(sdl3_name, hint_callback_wrapper, callback_info);
         SDL3_SetPointerProperty(hint_callbacks, property_name, NULL);
-        SDL_free(property_name);
+        SDL3_free(property_name);
     } else {
         SDL3_RemoveHintCallback(sdl3_name, callback, userdata);
     }
```

PR #412, opened 2025-03-15
tests: port failing SDL_Hint tests from pysdl2 to testautomation

madebr:


src/sdl2_compat.c
```diff
@@ -6225,12 +6226,32 @@ SDL_QueryTexture(SDL_Texture *texture, Uint32 *format, int *access, int *w, int
 SDL_DECLSPEC int SDLCALL
 SDL_SetTextureScaleMode(SDL_Texture *texture, SDL_ScaleMode scaleMode)
 {
+    SDL_PropertiesID texture_properties = SDL3_GetTextureProperties(texture);
+    SDL3_SetNumberProperty(texture_properties, PROP_TEXTURE_SCALE_MODE, scaleMode);
```
slouken, line 6230: It should probably only do this if setting the scale mode succeeds.

src/sdl2_compat.c
```diff
@@ -6225,12 +6226,32 @@ SDL_QueryTexture(SDL_Texture *texture, Uint32 *format, int *access, int *w, int
...
 
+SDL_DECLSPEC int SDLCALL
+SDL_GetTextureScaleMode(SDL_Texture *texture, SDL_ScaleMode *scaleMode)
+{
+    SDL_PropertiesID texture_properties = SDL3_GetTextureProperties(texture);
+    int cached_scale_mode;
+    if (texture_properties == 0) {
+        return 0;
+    }
+    cached_scale_mode = SDL3_GetNumberProperty(texture_properties, PROP_TEXTURE_SCALE_MODE, -99);
```
slouken, line 6245: Does this need a cast to int?

slouken: Also, I suggest `-1` as an invalid number.

src/sdl2_compat.c
madebr: I'm thinking to replace the properties with a linked list.
There's currently a leak in SDL_quit when not all hooked callbacks are removed.

madebr:
obsoleted by https://github.com/libsdl-org/sdl2-compat/pull/422

PR #421, opened 2025-03-17
quit: Don't unload SDL3 if SDL_Quit() wasn't called

### LINKED ISSUE #419: Crash when quitting in OpenAL threaded sound in LÖVE
Hey, I am running into a crash in Balatro when quitting the game (not critical, but interesting). I investigated for a while, and I found a minimal reproduction within LÖVE:
```lua
-- example/main.lua
local threadCode = [[
    require("love.sound")
    require("love.audio")
    require("love.timer")
    
    local musicPath = ...
    
    local music = love.audio.newSource(musicPath, "stream")
    music:setLooping(true)
    music:play()
    
    while true do
        love.timer.sleep(0.1)
    end
]]

function love.load()
    musicThread = love.thread.newThread(threadCode)
    musicThread:start("music.mp3")
end

function love.keypressed(key)
    if key == "escape" then
        love.event.quit()
    end
end
```
To run locally copy the above to `example/main.lua` and put in your favourite music to `example/music.mp3`. Repeatedly run `love example/` then press escape to quit at random intervals. After quitting multiple times you should get a crash with a backtrace like this:

```c++
#0  0x0000000000000000 in ?? ()
#1  0x00007a792cdca10e in SDL_LockMutex_REAL (a=<optimized out>) at /usr/src/debug/sdl2-compat/sdl2-compat-2.32.52/src/sdl2_compat.c:5818
#2  0x00007a792d081ba9 in love::thread::sdl::Mutex::lock (this=0x57588387c890) at modules/thread/sdl/threads.cpp:43
#3  love::thread::Lock::Lock (m=0x57588387c890, this=<synthetic pointer>) at modules/thread/threads.cpp:35
#4  love::audio::openal::Audio::PoolThread::threadFunction (this=0x57588385b280) at modules/audio/openal/Audio.cpp:57
#5  0x00007a792d152d4b in love::thread::sdl::Thread::thread_runner (data=0x5758838652f0) at modules/thread/sdl/Thread.cpp:95
#6  0x00007a792c1af357 in SDL_RunThread (thread=0x57588385b0c0) at /usr/src/debug/sdl3/SDL3-3.2.8/src/thread/SDL_thread.c:328
#7  RunThread (data=0x57588385b0c0) at /usr/src/debug/sdl3/SDL3-3.2.8/src/thread/pthread/SDL_systhread.c:73
#8  0x00007a792cea370a in start_thread (arg=<optimized out>) at pthread_create.c:448
#9  0x00007a792cf27aac in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

Not sure if this is an sdl2-compat, sdl3, love or OpenAL bug though unfortunately. I have tried to check what's happening in gdb, but the lock is opaque and also optimised out so that was not too helpful.

Tested on ArchLinux:
love 11.5-1
sdl2-compat 2.32.52-1
sdl3 3.2.8-1

cgutman:
This is a workaround for applications that don't synchronize threads using SDL2 with the exit handler that invokes the sdl2-compat library destructor.

Unfortunately, this may further suppress LeakSanitizer reports for developers using sdl2-compat if they forget to call `SDL_Quit()`, but this is preferable to crashing real apps that users care about.

Fixes #419

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -220,6 +220,7 @@ do { \
 
 
 static bool WantDebugLogging = false;
+static SDL_InitState InitSDL2CompatGlobals;
 
 
 static char *
@@ -977,6 +978,17 @@ static void dllinit(void)
 static void dllquit(void) __attribute__((destructor));
 static void dllquit(void)
 {
+    /* Some misbehaving applications may not join threads calling SDL2 functions
+       before exit handlers run and invoke the library destructor, which unloads
+       SDL3. As a workaround to avoid crashing those applications, we will skip
+       unloading SDL3 if SDL was used but SDL_Quit() was never called. */
+    if (SDL3_ShouldQuit(&InitSDL2CompatGlobals)) {
+        if (WantDebugLogging) {
+            SDL2Compat_LogAtStartup("sdl2-compat: Leaking SDL3 library reference due to missing call to SDL_Quit()");
+        }
+        return;
+    }
+
     UnloadSDL3();
 }
 
@@ -1089,8 +1101,6 @@ static SDL_SensorID SensorID2to3(SDL2_SensorID id);
 static SDL2_SensorID SensorID3to2(SDL_SensorID id);
 
 /* Functions! */
-
-static SDL_InitState InitSDL2CompatGlobals;
 
 /**
  * Verbosity of logged events as defined in SDL_HINT_EVENT_LOGGING:


sezero:
This is some really unfortunate patch. And how is windows supposed to be fixed, assuming it is affected as well?

cgutman:
Windows isn't affected because all running threads are forcefully terminated before `DLL_PROCESS_DETACH` is sent to loaded DLLs during process termination. Raymond Chen has a [blogpost](https://devblogs.microsoft.com/oldnewthing/20070503-00/?p=27003) about this and I also confirmed it in a debugger on Windows 11 24H2.

slouken:
Merged, thanks!

PR #428, opened 2025-03-24
Prefer X11 for the UE5 editor by default

### LINKED ISSUE #426: Mouse input not working in UE5 & Wayland
Using latest commit `f621c21c14fd0ecc144a13ee3741977edad18fcf`

Launched Unreal Engine 5.5.4 which internally uses SDL2 with Wayland support with sdl2-compat using `SDL_DYNAMIC_API`.
Windows renders correctly, but mouse input does not work. Keyboards does work. (for example, Tabs for navigation)

Interestingly cursor always shows resize, but it does not do anything including resize.
Image

`SDL_VIDEODRIVER=x11` works as intended. Only tested in XWayland which makes it buggy

Using Arch Linux with
sway 1:1.10.1-2
wlroots 0.18.2-1
wayland 1.23.1-1
nvidia 570.124.04-4

Kontrabant:
It requires more work before it will work properly on Wayland, as toast popup notifications are broken, some desktops have mouse input issues, and the current Nvidia drivers seem to have stability issues with it when running natively under Wayland (XWayland is fine, though).

Too many issues for now; favor X11/XWayland by default.

Fixes #426

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -492,6 +492,9 @@ static QuirkEntryType quirks[] = {
 
     /* Stylus Labs Write does its own X11 input handling */
     { "Write", "SDL_VIDEO_X11_XINPUT2", "0" },
+
+    /* The UE5 editor has input issues and broken toast notification positioning on Wayland */
+    { "UnrealEditor", SDL_HINT_VIDEO_DRIVER, "x11" },
 };
 
 #ifdef __linux__


Kontrabant:
Is setting the video driver hint in this manner the best way to go about it, or should a method that sets the `force_x11` parameter be used?

slouken:
This is fine. It's clear, self-documenting, and can be overridden by the user.

slouken (approved):

PR #435, opened 2025-03-26
avoid some integral constant will be truncated warnings

sezero:
gcc7 and gcc14 asm outputs are the same before and after this patch


slouken (approved):

sezero:
Pushed https://github.com/libsdl-org/sdl2-compat/commit/fd6e9453c24a1d093b89955582ff73efa12e0294 as a follow-up for consistency. gcc7.5 (tested mingw 32 and 64 bits and linux 32 biits) outputs the same asm. gcc14 output is not identical, though.

PR #446, opened 2025-04-01
Rename sdl2-compat.pc to sdl2.pc

slouken:
sdl2-compat should be a drop-in replacement for SDL2, so use the same pkg-config package name

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -599,8 +599,8 @@ if(SDL2COMPAT_INSTALL)
   set(ENABLE_SHARED_TRUE "")
   set(ENABLE_SHARED_FALSE "#")
 
-  configure_file(sdl2-compat.pc.in sdl2-compat.pc @ONLY)
-  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sdl2-compat.pc"
+  configure_file(sdl2.pc.in sdl2.pc @ONLY)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sdl2.pc"
     DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig
   )
 
--- a/cmake/test/test_pkgconfig.sh
+++ b/cmake/test/test_pkgconfig.sh
@@ -30,10 +30,10 @@ fi
 
 # Get the canonical path of the folder containing this script
 testdir=$(cd -P -- "$(dirname -- "$0")" && printf '%s\n' "$(pwd -P)")
-SDL_CFLAGS="$( pkg-config sdl2-compat --cflags )"
-SDL_LDFLAGS="$( pkg-config sdl2-compat --libs )"
+SDL_CFLAGS="$( pkg-config sdl2 --cflags )"
+SDL_LDFLAGS="$( pkg-config sdl2 --libs )"
 if [ "x$test_static" = "xyes" ]; then
-    SDL_STATIC_LDFLAGS="$( pkg-config sdl2-compat --libs --static )"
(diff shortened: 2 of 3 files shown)


sezero:
Same thing is true for sdl12-compat

slouken:
> Same thing is true for sdl12-compat

Yep, you want to fix that?

Conan-Kudo:
Why was this done? It's not actually needed, especially since pkgconf reads the `Provides:` line just fine and expresses it with dependencies. That's why both `sdl12-compat` and `sdl2-compat` followed this strategy.

Conan-Kudo:
Today, all major distributions (Arch, Debian, Fedora, openSUSE) and minor ones like Alpine use pkgconf's implementation. It's also the main implementation used in Windows SDKs now too because it eliminates an annoying bootstrap problem. This change was unnecessary and also wrong.

Conan-Kudo:
> > Why was this done? It's not actually needed, especially since pkgconf reads the Provides: line just fine and expresses it with dependencies.
> 
> Some tools unfortunately don't recognise `Provides:`. Obviously that's their problem really, but it's bad enough that both [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/sdl2-compat/-/merge_requests/2) and [Debian](https://salsa.debian.org/sdl-team/sdl2-compat/-/blob/debian/latest/debian/libsdl2-compat-dev.links) insert symlinks in their packaging.

It isn't needed for either of them now in the releases that it would be shipped in, since both use `pkgconf` (which does support this).

georgefst:
I know `pkgconf` supports it. What I mean is that there are tools which call `pkgconf` in such a way that `Provides:` is ignored. For example, https://github.com/haskell/cabal/issues/10808.

Conan-Kudo:
Let's fix those things then, because we aren't the only project doing this.

georgefst:
> Let's fix those things then

Yeah, absolutely, we should!

I'm not even sure that merging this PR was the right call. I'm just trying to explain what was presumably the rationale.

keithbowes:
> > > Why was this done? It's not actually needed, especially since pkgconf reads the Provides: line just fine and expresses it with dependencies.
> > 
> > 
> > Some tools unfortunately don't recognise `Provides:`. Obviously that's their problem really, but it's bad enough that both [Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/sdl2-compat/-/merge_requests/2) and [Debian](https://salsa.debian.org/sdl-team/sdl2-compat/-/blob/debian/latest/debian/libsdl2-compat-dev.links) insert symlinks in their packaging.
> 
> It isn't needed for either of them now in the releases that it would be shipped in, since both use `pkgconf` (which does support this).

Well, `pkgconf --exists sdl2` succeeds, but `pkgconf --atleast-version 2.0 sdl2` fails (same with - -exact-version and --max-version), thus trying to compile software that checks whether a required version of SDL2 is installed will fail without a symlink present.

Conan-Kudo:
> 
> Well, `pkgconf --exists sdl2` succeeds, but `pkgconf --atleast-version 2.0 sdl2` fails (same with - -exact-version and --max-version), thus trying to compile software that checks whether a required version of SDL2 is installed will fail without a symlink present.

This seems like a bug, because I don't think those are supposed to fail.

slouken:
> I'm not even sure that merging this PR was the right call. I'm just trying to explain what was presumably the rationale.

FYI, we backed out this PR in https://github.com/libsdl-org/sdl2-compat/commit/51a1c8271262fb1d89affe0c61f3c7c3f112bcd0

Issue #313, opened 2025-02-04
Object destruction issues...


icculus:
Something I'm seeing in a lot of the apps we've fixed things in over the last several days:

- They call SDL_Quit() when cleaning up.
- They call exit() or return from main().
- They then call SDL_FreeSurface or SDL_FreeAudioStream, etc, during static destructors.

SDL3 already destroyed these things during SDL_Quit, which I don't think SDL2 does, so when the static destructors run, they're already gone.

Here's The Battle For Wesnoth, just starting and clicking the window bar's close button on the main menu:

```
corrupted size vs. prev_size while consolidating

Thread 1 "wesnoth" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
warning: 44	./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff6f7326e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff6f568ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff6f577b6 in __libc_message_impl (fmt=fmt@entry=0x7ffff70fc8d7 "%s\n") at ../sysdeps/posix/libc_fatal.c:132
#6  0x00007ffff6fd6fe5 in malloc_printerr (str=str@entry=0x7ffff70ffb30 "corrupted size vs. prev_size while consolidating") at ./malloc/malloc.c:5772
#7  0x00007ffff6fd9144 in _int_free_merge_chunk (av=0x7ffff7131ac0 <main_arena>, p=0x5555576ec380, size=14734880) at ./malloc/malloc.c:4695
#8  0x00007ffff6fd942a in _int_free (av=0x7ffff7131ac0 <main_arena>, p=<optimized out>, have_lock=<optimized out>) at ./malloc/malloc.c:4646
[...]

PR #450, opened 2025-04-06
surface: Rework lifetime management to use SDL3 property cleanup

### LINKED ISSUE #449: SDL_LockTextureToSurface leaks memory
Don't have a minimal example for this but a relevant snippet is: 
```c
    SDL_Surface *tmp_surf;
    SDL_LockTextureToSurface(window_tex, NULL, &tmp_surf);
    SDL_BlitSurface((SDL_Surface *)PaletteSurface, NULL, tmp_surf, NULL);
    SDL_UnlockTexture(window_tex);
```
(https://github.com/Daft-Freak/CnC_and_Red_Alert/blob/main/SDLLIB/drawbuff_sdl.cpp#L69-L72)

Doesn't seem like the surface wrapper gets freed on unlock.

cgutman:
This PR changes the lifecycle management of SDL2 surfaces to be tied to the SDL3 surface lifetime via property cleanup callbacks. The sole exception is the SDL2 window surface which remains associated with the window to ensure apps that erroneously pass the window surface to `SDL_FreeSurface()` don't crash.

Fixes #449

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -3811,6 +3811,21 @@ SDL_LoadWAV_RW(SDL2_RWops *rwops2, int freesrc, SDL2_AudioSpec *spec2, Uint8 **a
     return retval;
 }
 
+static void FreeSurface2(SDL2_Surface *surface)
+{
+    if (surface->format) {
+        SDL_SetPixelFormatPalette(surface->format, NULL);
+        SDL_FreeFormat(surface->format);
+    }
+
+    SDL3_free(surface);
+}
+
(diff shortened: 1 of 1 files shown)


slouken:
Life cycle management was originally implemented like this, but caused problems for applications that used surfaces after SDL_Quit(), e.g. the py-sdl2 gui.py example as mentioned in https://github.com/libsdl-org/sdl2-compat/issues/313. It seems like this change would reintroduce those problems, or am I missing something?

cgutman:
Yep, you're correct that this would likely reintroduce issues in #313. I'll abandon this in favor of #453 which mimics your window surface approach.

PR #451, opened 2025-04-06
Revert "Rename sdl2-compat.pc to sdl2.pc (#446)"

Conan-Kudo:
This is unnecessary, given that everyone uses `pkgconf(1)`, which fully supports the `Provides:` stanza and expresses it in dependencies.

The issue mentioned in the commit being reverted is a matter of education.

This reverts commit 0bc8d4434fb911973762230bbcf4529f39fb8912.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -599,8 +599,8 @@ if(SDL2COMPAT_INSTALL)
   set(ENABLE_SHARED_TRUE "")
   set(ENABLE_SHARED_FALSE "#")
 
-  configure_file(sdl2.pc.in sdl2.pc @ONLY)
-  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sdl2.pc"
+  configure_file(sdl2-compat.pc.in sdl2-compat.pc @ONLY)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/sdl2-compat.pc"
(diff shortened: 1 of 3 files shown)


sezero:
> This is unnecessary

Does it really break anything, though?

smcv:
> > This is unnecessary
> 
> Does it really break anything, though?

I think this is the key question here.

Is there any situation where a dependent project would be correct to ask the dependency management for `sdl2-compat` or `sdl2_compat` specifically, as opposed to asking for `sdl2` and then getting either classic SDL2 or sdl2-compat (whichever one happens to be installed) in response?

> everyone uses `pkgconf(1)`

Backports into older distributions don't necessarily. The Steam Runtime has actively-maintained backports of sdl2-compat into branches based on Debian 10 and Debian 11, which used the reference (freedesktop.org) pkg-config, not pkgconf.

(At the moment the Steam Runtime SDK's SDL2 development files come from classic SDL2, and classic SDL2 is also the default at runtime, but this is likely to swap to sdl2-compat when SDL3 and sdl2-compat are more mature.)

slouken:
Okay, after discussion we'll go ahead and accept this PR. Thanks for the feedback!

Conan-Kudo:
Thank you very much for doing so! 😄

slouken:
You're welcome!

PR #453, opened 2025-04-08
texture: Fix leak in SDL_LockTextureToSurface()

### LINKED ISSUE #449: SDL_LockTextureToSurface leaks memory

cgutman:
Use the same surface lifetime management in `SDL_LockTextureToSurface()` as we do for `SDL_GetWindowSurface()` (except the lifetime of the SDL2 surface is tied to the SDL3 surface instead of the window).

Since this surface is not supposed to be freed via `SDL_FreeSurface()`, we shouldn't have the same lifetime issues with it as with regular surfaces in #313.

Fixes #449

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -9971,7 +9971,8 @@ SDL_GetSurfaceBlendMode(SDL2_Surface *surface, SDL_BlendMode *blendMode)
     return SDL3_GetSurfaceBlendMode(Surface2to3(surface), blendMode) ? 0 : -1;
 }
 
-static void SDLCALL CleanupWindowSurface(void *userdata, void *value)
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #455, opened 2025-04-08
Don't use quirks to force X11 things on non-X11 platforms

smcv:
Followup for #454.

* Centralize detection of platforms that probably support X11
* Don't use quirks to force X11 things on non-X11 platforms

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -505,8 +505,10 @@ static QuirkEntryType quirks[] = {
     { "tauon/__main__.py", SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "0" },
     { "tauon/__main__.py", SDL_HINT_VIDEO_WAYLAND_ALLOW_LIBDECOR, "0" },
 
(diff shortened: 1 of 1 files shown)


smcv:
In principle we could have a similar thing for "platforms where Wayland might be relevant", although I'm not 100% sure what that would mean. (Only Linux? Any Unix other than Android? Any Unix other than Android and Apple?)

src/sdl2_compat.c
```diff
@@ -111,6 +111,13 @@ This breaks the build when creating SDL_ ## DisableScreenSaver
 #define SDL2COMPAT_MAXPATH 1024
 #endif
 
+#if defined(SDL_PLATFORM_UNIX) && !defined(SDL_PLATFORM_ANDROID)
```
sezero, line 114: Android w/ x11 ?

sezero: Ugh. ignore me, failed to notice that `!`

slouken:
Merged, thanks!

PR #462, opened 2025-04-13
mouse: Reintroduce fractional mouse wheel handling with fixes

### LINKED ISSUE #461: OpenTTD relies on both fractional wheel values and accumulated integer values
OpenTTD relies on both fractional values (for smooth two-finger viewport dragging) and accumulated integer values (for scrolling in menus). Should it be quirked?

_Originally posted by @v1993 in https://github.com/libsdl-org/sdl2-compat/issues/403#issuecomment-2769844932_

cgutman:
The approach of having SDL3 perform integer accumulation for wheel events doesn't work for apps that use both fractional and integer scrolling. This PR brings scroll wheel accumulation back into sdl2-compat.

The previous version of integer accumulation in #378 was actually broken because it did not account for the fact that a single SDL3 event can be converted using `Event3to2()` multiple times. If that happened (ex: due to an event filter/watcher installed by the app), accumulation of residual values would happen faster than expected. This PR solves that by introducing a context struct that is passed to `Event3to2()` and reset as necessary to ensure each SDL3 event conversion results in the same SDL2 event values in each context where event conversion occurs.

While this approach does solve the problem, I don't particularly like all the context passing and reverting all over. I'm hoping someone has a better idea to acheive idempotence for `Event3to2()`. I've left this as draft for now until we decide if this approach is actually what we want to land.

Fixes #461

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -488,12 +488,9 @@ static QuirkEntryType quirks[] = {
     { "hl.exe", SDL_HINT_MOUSE_EMULATE_WARP_WITH_RELATIVE, "0" },
 #endif
 
-    /* Moonlight supports high DPI properly under Wayland.
-       It also reads fractional values in wheel events. */
+    /* Moonlight supports high DPI properly under Wayland */
     { "moonlight", SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "0" },
     { "moonlight-qt", SDL_HINT_VIDEO_WAYLAND_SCALE_TO_DISPLAY, "0" },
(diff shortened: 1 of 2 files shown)


slouken:
Yeah, this is not great. What does it mean to use both fractional and integer scrolling?

slouken:
Let's just add integer fields back into the SDL3 event and have it do the accumulation there. That should be safe to do for 3.2.12, if you want to take a quick crack at it.

cgutman:
Yep, will do

cgutman:
Superseded by #480 which uses the new SDL3 fields

Issue #457, opened 2025-04-12
Output borken in OpenXRay with Wayland driver


OlegAckbar:
When using OpenXRay with sdl2-compat and Wayland driver output is broken. No such issue on X11 driver or SDL2 with Wayland driver.

https://github.com/user-attachments/assets/89c2855d-b923-4dd1-bc25-449e56135f85

Kontrabant:
Just glancing at the OpenXRay code, this might be related to #456.

OlegAckbar:
@Kontrabant no, it didn't fix the issue

Kontrabant:
Could you describe your desktop layout (position, resolution, and scale factor of each display)?

OlegAckbar:
![Image](https://github.com/user-attachments/assets/bf4d0393-8989-4141-b31e-9cee1de9c9e1)

Kontrabant:
The issue here is that fullscreen mode is being set before the window is mapped, and the compositor is ignoring the requested output in favor of initially positioning the window itself in what it probably deems to be a better location (it seems that both GNOME and KDE do this). With this desktop layout replicated, I can reproduce this with more than OpenXRay (vkquake and others also demonstrate the issue).

Fullscreen windows will need their positions checked and corrected after the window is mapped in SDL3.

Kontrabant:
Looking further into this, it seems that the GNOME positioning issue is a separate matter (https://gitlab.gnome.org/GNOME/mutter/-/issues/3854). The issue with OpenXRay is as follows:
- Creates a 640x480 window at 0,0
- Queries the primary display
- Moves the window onto the primary display
- Enters fullscreen

sdl2-compat queries the current window position when setting fullscreen, however, as Wayland can't move the window, the window is being reported as still being on the display with 0,0 coordinates. SDL3 has internal logic to handle this when selecting a fullscreen display, but it's not accessible via the public API, so another way of doing this is required.

PR #463, opened 2025-04-16
video: Allow SDL_SetWindowSize to change the exclusive fullscreen mode

### LINKED ISSUE #456: SetWindowsSize doesn't update the display mode when a game is already fullscreen
In SDL2 if you call `SDL_SetWindowSize` while a window is already Exclusive Fullscreen it will update the resolution to try and match the window size if you never set anything via `SDL_SetWindowDisplayMode`.

In SDL3 & sdl2-compat it ignores `SDL_SetWindowSize` while in Fullscreen. (from what I can see)

I noticed in the MonoGame source code this is how they do it for Exclusive Fullscreen.
```
SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN);
SDL_SetWindowSize(window, width, height);
```

So just mainly opening this issue from a compatibility perspective, but I will possibly open a PR for MonoGame since I am assuming their way is trying to change the display mode twice.
- Once trying to match the old window size when `SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN);` is called
- Again when trying to match the new window size from `SDL_SetWindowSize(window, width, height);`

Kontrabant:
Replicate the behavior of SDL2, which allows changing the fullscreen mode by changing the window size while the window is exclusive fullscreen.

Additionally, changing the size of an exclusive fullscreen window also affects the size of the window when coming out of fullscreen in SDL2.

Fixes #456 (maybe #457 too?)

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -8440,12 +8440,15 @@ SDL_GetDesktopDisplayMode(int displayIndex, SDL2_DisplayMode *mode)
 }
 
 #define PROP_WINDOW_FULLSCREEN_MODE "sdl2-compat.window.fullscreen-mode"
+#define PROP_WINDOW_FULLSCREEN_RESIZE_W "sdl2-compat.window.fullscreen_resize_w"
+#define PROP_WINDOW_FULLSCREEN_RESIZE_H "sdl2-compat.window.fullscreen_resize_h"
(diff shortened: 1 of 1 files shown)


arosolino:
Good stuff @Kontrabant !

PR #472, opened 2025-04-29
surface: Fix crash when unlocking an RLE surface

### LINKED ISSUE #471: Double free when working with RLE surfaces
While trying to migrate pygame-ce form sdl2 classic to compat I've run into double free errors in its tests. After some debugging I've managed narrow it down to `test_fill_rle` test. Here's the C version for easier debugging:

```c
#include <SDL2/SDL.h>

int main() {
  SDL_Surface *surf = SDL_CreateRGBSurface(0, 32, 32, 32, 0, 0, 0, 0);
  SDL_Surface *blit_surf = SDL_CreateRGBSurface(0, 32, 32, 32, 0, 0, 0, 0);

  SDL_SetSurfaceRLE(blit_surf, 1);
  SDL_SetColorKey(blit_surf, 1, 0xff00ff);

  SDL_BlitSurface(blit_surf, NULL, surf, NULL);

  SDL_LockSurface(blit_surf);
  SDL_FillRect(blit_surf, NULL, 0xfaf9f9ff);
  SDL_UnlockSurface(blit_surf);
}
```

This code runs fine with sdl2 classic but erros with sdl2-compat (tested with `2.32.54` and latest `main`).

```sh
$ gcc -o main main.c `pkg-config --cflags --libs sdl2`
$ ./main 
$ LD_LIBRARY_PATH=~/repos/sdl2-compat/build ./main 
double free or corruption (out)
Aborted (core dumped)
```

The same code runs fine when using analogous functions from sdl3 so I suspect that there is some issue with the functions converting between SDL2 and SDL3 surfaces.

cc https://github.com/pygame-community/pygame-ce/issues/3287

cgutman:
The surface may be reallocated with SIMD alignment by SDL3, so we synchronize the flags with the SDL2 surface to indicate that.

Fixes #471

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -9948,8 +9948,7 @@ SDL_LockSurface(SDL2_Surface *surface)
         if (!SDL3_LockSurface(surface3)) {
             return -1;
         }
-        surface->pixels = surface3->pixels;
-        surface->pitch = surface3->pitch;
+        SynchronizeSurface3to2(surface3, surface);
     }
 
     /* Increment the surface lock count, for recursive locks */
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #479, opened 2025-05-01
window: Fix SDL_GetWindowFlags(NULL) behavior to match SDL2

### LINKED ISSUE #477: `SDL_GetWindowFlags(NULL)` doesn't return 0
This was caught by `test_get_active` test in pygame-ce repository.

SDL3 classic performs `CHECK_WINDOW_MAGIC(window, 0);` when calling `SDL_GetWindowFlags` making it return 0 if the provided window is not valid. Since SDL2 had redundant flags, valid windows always had at least one flag set – either `SDL_WINDOW_HIDDEN` or `SDL_WINDOW_SHOWN`. SDL3 also performs this check, however, since the redundant flags were removed, 0 is no longer a trap representation but one of the possible values for a valid window. sdl2-compat assumes that this is always the case thus returning non zero flags given an invalid window.

This could be fixed by either exposing some way of checking window validity in SDL3 or by adding some dummy flag like `SDL_WINDOW_VALID` that is always set for valid windows thus making 0 into a trap representation in SDL3 as well.

cc https://github.com/pygame-community/pygame-ce/issues/3287

cgutman:
`SDL_GetWindowID()` is a cheap way to identify a valid window with no side-effects.

Fixes #477


slouken:
Merged, thanks!

PR #484, opened 2025-05-07
Make X11 dependency required on UNIX platforms

marcin-serwin:
The X11 headers are included in `sdl2_compat.h` so building without them installed in the standard location fails with:

```
> In file included from /build/source/src/dynapi/SDL_dynapi.c:23:
> /build/source/src/dynapi/../sdl2_compat.h:1289:10: fatal error: X11/Xlib.h: No such file or directory
>  1289 | #include <X11/Xlib.h>
>       |          ^~~~~~~~~~~~
> compilation terminated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,15 +223,16 @@ set(SDL2COMPAT_SRCS
   src/dynapi/SDL_dynapi.c
 )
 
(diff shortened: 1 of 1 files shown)


slouken:
Merged, thanks!

PR #490, opened 2025-05-20
Update palette in Surface2to3

sulix:
As with some other surface properties, it's possible that an application will modify a surface's palette behind SDL2's back.  In particular, this is often used as a shortcut to emulate DirectDraw's blitting behaviour (which ignores the palette).

Synchronise the palette in Surface2to3 alongside other properties which are modified in this manner. _I think_ it should be safe to use SDL3_SetSurfacePalette() here, as if the surface's palette is unchanged, the refcount should also end up unchanged.


slouken:
Looks good, thanks!

Issue #515, opened 2025-09-10
should log a message if SDL_DYNAMIC_API results in loading classic SDL


smcv:
To help Debian to migrate from classic SDL2 to sdl2-compat, I've proposed a packaging change that will give us a `libsdl2-classic` package, installed outside the default search path. That allows users to have one of four configurations:

1. install `libsdl2-2.0-0` (now depends on `libsdl2-classic`): classic SDL2 only
2. install `libsdl2-2.0-0` (now depends on `libsdl2-classic`) and `libsdl2-compat`: classic SDL2 by default, sdl2-compat opt-in via `export SDL_DYNAMIC_API='/usr/$LIB/sdl2-compat/libSDL2-2.0.so.0'` or `export LD_LIBRARY_PATH='/usr/$LIB/sdl2-compat'`
3. install `libsdl2-compat-shim` (depends on `libsdl2-compat`) and `libsdl2-classic`: sdl2-compat by default, classic SDL2 opt-in via `export SDL_DYNAMIC_API='/usr/$LIB/sdl2-classic/libSDL2-2.0.so.0'` or `export LD_LIBRARY_PATH='/usr/$LIB/sdl2-classic'`
4. install `libsdl2-compat-shim` (depends on `libsdl2-compat`): sdl2-compat only

The idea is that when sdl2-compat has had more testing, we can flip the defaults around so that `libsdl2-2.0-0` is provided by and contains the sdl2-compat library, and `libsdl2-classic` (the equivalent of configuration 3) is the only way to get classic SDL2.

However, on a system with configuration 3, when I run

```
SDL2COMPAT_DEBUG_LOGGING=1 \
SDL_DYNAMIC_API='/usr/$LIB/sdl2-classic/libSDL2-2.0.so.0' \
openarena
```

I get this output, the same as I would have done in configuration 2 or 4:

```
sdl2-compat 2.32.56, built on [...], talking to SDL3 3.3.0
sdl2-compat: This app appears to be named: ioquake3
...
```

which doesn't give any indication that sdl2-compat has been overridden/replaced by classic SDL2.

I think it would be good if a successful use of `SDL_DYNAMIC_API` under `SDL2COMPAT_DEBUG_LOGGING=1` logged an additional message, perhaps something like:

```
SDL_DYNAMIC_API: sdl2-compat replaced by SDL 2.32.10 (SDL-release-2.32.10-0-g5d2495703 (Debian 2.32.10+dfsg-3))
```

[...]

PR #516, opened 2025-09-10
dynapi: Under SDL2COMPAT_DEBUG_LOGGING, log whether we were replaced

smcv:
Otherwise our output can be misleading: we say that sdl2-compat is loaded and using SDL3 (because that happens at library constructor time, before any SDL2 API functions are called), but then SDL_DYNAMIC_API kicks in the first time we call into the SDL2 API, and potentially replaces us with classic SDL2 without leaving a trace in our output.

Resolves: https://github.com/libsdl-org/sdl2-compat/issues/515

---

This is one possible implementation of #515. I've tried to write it in a style that could be copied into SDL3, or even into classic SDL2, if wanted.

Example output:

```
sdl2-compat 2.32.57, built on Sep 10 2025 at 12:31:43, talking to SDL3 3.3.0
sdl2-compat: This app appears to be named: ioquake3
SDL_DYNAMIC_API: Library replaced by SDL 2.32.10 (SDL-release-2.32.10-0-g5d2495703 (Debian 2.32.10+dfsg-3))
```

or

```
sdl2-compat 2.32.57, built on Sep 10 2025 at 12:31:43, talking to SDL3 3.3.0
sdl2-compat: This app appears to be named: ioquake3
SDL_DYNAMIC_API: Not overridden
```

### DIFF
--- a/src/dynapi/SDL_dynapi.c
+++ b/src/dynapi/SDL_dynapi.c
@@ -445,7 +445,7 @@ DynApiExitProcess(int exitcode)
 }
 
 
-static void SDL_InitDynamicAPILocked(void)
+static bool SDL_InitDynamicAPILocked(void)
 {
     SDL_DYNAPI_ENTRYFN entry = NULL; /* funcs from here by default. */
     bool use_internal = true;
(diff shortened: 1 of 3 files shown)


smcv:
> SDL_DYNAMIC_API: Not overridden

I'm unsure whether it's better to log this (to say "yes, you're using a version of sdl2-compat that would tell you if it had been overridden, and no, it wasn't") or remain silent, so I erred on the side of being more informative.

sezero:
>  @sezero, do you see any issues here?

None as far as I can see. @icculus?

PR #523, opened 2025-09-24
Add a new quirk 'SDL2COMPAT_NO_UNLOAD', usable/needed by SC3U

twolife:
Just like in https://github.com/libsdl-org/sdl12-compat/pull/360
When chaining sdl12-compat -> sdl2-compat -> SDL3 , that 'hack' needs to be present in both shim


slouken:
Merged, thanks!

PR #525, opened 2025-09-30
Treat DEBUG_INVOCATION=1 as equivalent to SDL2COMPAT_DEBUG_LOGGING=1

smcv:
In the same spirit as https://github.com/libsdl-org/SDL/issues/12275, if we enable a reasonable level of debug output whenever this variable is set, it becomes a convenient way for users to request a more verbose (but not overwhelming) amount of output. SDL 3 and GLib already do this.

Explicitly setting SDL2COMPAT_DEBUG_LOGGING=0 can override this, if it becomes necessary for whatever reason.


slouken:
Merged, thanks!

slouken:
For style let's also switch to:
```
if (!value) {
...
```

slouken:
I'll go ahead and fix this if you're not already doing it?

PR #543, opened 2025-11-04
Preserve window creation errors set by SDL 3

weirddan455:
I was trying to debug a window creation error in my application but `SDL_GetError()` was returning the unhelpful error "Invalid window" because sdl2_comapt is trying to do things to an invalid window.

I added an early check to skip these functions if SDL3's window creation returns NULL. This way we don't stomp on the error message.


slouken:
Merged, thanks!

PR #547, opened 2025-11-07
Fix uninitialized variable warning when compiling tests with clang compiler

Sackzement:
Compiling the tests with `Clang` shows these warnings:
```
[ 92%] Building C object test/CMakeFiles/testautomation.dir/testautomation_rect.c.o
sdl2-compat/test/testautomation_rect.c:744:39: warning: variable 'rectA' is uninitialized when passed as a const
      pointer argument here [-Wuninitialized-const-pointer]
  744 |     intersection = SDL_IntersectRect(&rectA, (SDL_Rect *)NULL, &result);
      |                                       ^~~~~
sdl2-compat/test/testautomation_rect.c:998:41: warning: variable 'rectA' is uninitialized when passed as a const
      pointer argument here [-Wuninitialized-const-pointer]
  998 |     intersection = SDL_HasIntersection(&rectA, (SDL_Rect *)NULL);
      |                                         ^~~~~
sdl2-compat/test/testautomation_rect.c:1533:20: warning: variable 'rectA' is uninitialized when passed as a const
      pointer argument here [-Wuninitialized-const-pointer]
 1533 |     SDL_UnionRect(&rectA, (SDL_Rect *)NULL, &result);
      |                    ^~~~~
```

This commit initializes the rect variables (and makes them `const`).
It also gives them some size, so the tests only pass if a parameter is `NULL` and not accidentally on an empty rect.

Btw:
What is the reasoning behind the function returning `false` when the `result` parameter is `NULL`?
I expected it to return `true` regardless if there is a parameter to write the resulting rect to, similar to `SDL_HasRectIntersection()`.

### DIFF
--- a/test/testautomation_rect.c
+++ b/test/testautomation_rect.c
@@ -733,8 +733,8 @@ int rect_testIntersectRectEmpty(void *arg)
  */
 int rect_testIntersectRectParam(void *arg)
 {
-    SDL_Rect rectA;
-    SDL_Rect rectB = { 0 };
+    const SDL_Rect rectA = { 0, 0, 32, 32 };
+    const SDL_Rect rectB = { 0, 0, 32, 32 };
     SDL_Rect result;
     SDL_bool intersection;
 
@@ -988,8 +988,8 @@ int rect_testHasIntersectionEmpty(void *arg)
  */
 int rect_testHasIntersectionParam(void *arg)
 {
-    SDL_Rect rectA;
-    SDL_Rect rectB = { 0 };
+    const SDL_Rect rectA = { 0, 0, 32, 32 };
+    const SDL_Rect rectB = { 0, 0, 32, 32 };
     SDL_bool intersection;
 
     /* invalid parameter combinations */
@@ -1524,7 +1524,8 @@ int rect_testUnionRectInside(void *arg)
  */
 int rect_testUnionRectParam(void *arg)
 {
-    SDL_Rect rectA, rectB = { 0 };
+    const SDL_Rect rectA = { 0, 0, 32, 32 };
+    const SDL_Rect rectB = { 0, 0, 32, 32 };
     SDL_Rect result;
 
     /* invalid parameter combinations */


sezero:
> Compiling the tests with `Clang` shows these warnings:

Is clang silent in real SDL2? It would be better if this is resolved in real SDL2 and copied over here, unless it is sd2compat-specific.

> Btw: What is the reasoning behind the function returning `false` when the `result` parameter is `NULL`? I expected it to return `true` regardless if there is a parameter to write the resulting rect to, similar to `SDL_HasRectIntersection()`.

@slouken

Sackzement:
> Is clang silent in real SDL2? It would be better if this is resolved in real SDL2 and copied over here, unless it is sd2compat-specific.

It also shows up in SDL2:
```
[ 82%] Linking C executable testhittesting
SDL-2/test/testautomation_rect.c:744:39: error: variable 'rectA' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
  744 |     intersection = SDL_IntersectRect(&rectA, (SDL_Rect *)NULL, &result);
      |                                       ^~~~~
[ 82%] Built target testhittesting
SDL-2/test/testautomation_rect.c:998:41: error: variable 'rectA' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
  998 |     intersection = SDL_HasIntersection(&rectA, (SDL_Rect *)NULL);
      |                                         ^~~~~
SDL-2/test/testautomation_rect.c:1533:20: error: variable 'rectA' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
 1533 |     SDL_UnionRect(&rectA, (SDL_Rect *)NULL, &result);
      |                    ^~~~~
```

I will post a PR there too in a bit.

Edit: It's only an error here, because it's compiled with `-Werror` to catch it quickly.

Sackzement:
Here is the SDL2 PR: https://github.com/libsdl-org/SDL/pull/14416

PR #552, opened 2025-11-22
test: copy surface_testSetGetClipRect from SDL2

madebr:


madebr:
The test still needs to be ported to sdl3. I'll do that this weekend.

PR #559, opened 2025-12-19
Improve support for GNU/Hurd

pinotree:
SDL already supports GNU/Hurd, so improves the integration for it in this SDL2 compatibility layer:
- tighten/improve the platform detection in cmake, and use "Hurd" as identifier
- use the Linux `OS_GetExeName()` using `/proc/self/exe` as provided by the basic Linux-like procfs

This is the equivalent (much simpler of course) of https://github.com/libsdl-org/SDL/pull/13711; the RISCOS changes are kept as separate commit.


madebr:
This PR is technically correct, but I share sezero's point: it is a question of what this project wants to support. Y,he condition also does not contain any BSD.

A static sdl2-compat library would still require a shared sdl3 library, so the "advantages" of a static library are limited.

PR #560, opened 2025-12-19
tests: extend testatomic timeout to 120 seconds

pinotree:
The default (i.e. 10 seconds) is not enough for slower architectures, such as alpha or hppa.

Raise it to 120 seconds according to the feedback of hppa porters: https://bugs.debian.org/1095774


sezero:
If accepted, this should go into SDL2 and possibly into SDL3 too

madebr (approved):
120s for all platforms is not unreasonable

sezero:
> > SDL2 autotools doesn't have a `make check` functionality
> 
> It does, 

Ah, indeed, looks like it was added in 2.24.0. Sorry.

Issue #564, opened 2025-12-24
regression vs. "classic" SDL2: can't run setuid/setgid games


smcv:
As noted in https://github.com/libsdl-org/sdl12-compat/blob/main/HOW_TO_TEST_GAMES.md, some distributions make a subset of games setuid or setgid. A typical use is making them setgid to a group named `games` to be able to write out high score tables or similar records that are shared between all users of a multi-user system. (I personally think this is not a sufficiently important thing to justify use of setgid, but I didn't design this feature or set this policy.)

Because SDL 3 uses GTK to implement "tray icons", this results in the game triggering GTK's safety-catch against being run setuid/setgid, which the authors of GTK specifically do not support. For example in Debian testing/unstable:

```
$ apt install libsdl2-compat angband
$ SDL_DYNAMIC_API='/usr/$LIB/sdl2-compat/libSDL2-2.0.so.0' angband -msdl2; echo "(exit status $?)"

(process:624673): Gtk-WARNING **: 18:12:37.294: This process is currently running setuid or setgid.
This is not a supported use of GTK+. You must create a helper
program instead. For further details, see:

    http://www.gtk.org/setuid.html

Refusing to initialize GTK+.
(exit status 1)
```

Could SDL 3 perhaps have a hint that disables use of GTK for tray icons (which could either be phrased as "disable GTK" or "disable tray icons" whichever one you think is better), which sdl2-compat would set automatically? SDL 2 didn't have an API for tray icons *anyway*, so no functionality would be lost.

smcv:
I would be very surprised if either SDL 2 or SDL 3 is "secure" against being subverted to execute arbitrary code when loaded into a setgid process, so this use of setgid is security theatre at best, and I think GTK is likely doing the right thing here. But this is a user-observable regression.

smcv:
> Could SDL 3 perhaps have a hint that disables use of GTK for tray icons?

It already does, `SDL_ENABLE_GTK=0`, so sdl2-compat just needs to set that.

smcv:
As well as setting the hint (#565), to solve this class of issue it would be necessary to avoid libdecor loading its GTK plugin (libsdl-org/SDL#14712) or perhaps avoid libdecor entirely.

slouken:
I think we want to use the GTK plugin where possible, because it looks better and fits in more with users' desktops.

smcv:
[...]

PR #565, opened 2025-12-27
Disable GTK during initialization

smcv:
If SDL 3 loads and initializes GTK, GTK will exit the process if it detects a setgid executable. This is incompatible with the historical practice of making game executables setgid in order to write out a shared high-score table on multi-user systems (which is security theatre at best, because typical game runtime libraries are not hardened against an untrusted caller, but making it regress would be a user-observable regression in sdl2-compat).

Helps: https://github.com/libsdl-org/sdl2-compat/issues/564

---

Successfully tested with [Help Hannah's Horse](https://tracker.debian.org/pkg/hannah), but additionally requires https://github.com/libsdl-org/SDL/pull/14712 to be fully effective on Wayland desktops.


slouken:
As mentioned in the other bugs, I think we want to enable GTK by default, but disable libdecor's use of it (and probably print a debug level log message) when setuid/setgid in SDL3.

Issue #533, opened 2025-10-20
[Question] How to get SDL2COMPAT_DISABLE_X11 properly defined in SDL_config_unix.h?


vanfanel:
Good morning,

Recently, I added the `-DSDL2COMPAT_DISABLE_X11=ON` CMake option to be able to build SDL2 Compat in X11-less GNU/Linux systems, same as it was possible to build SDL2 on such systems. It works fine.

However, after building SDL2 Compat, `SDL2COMPAT_DISABLE_X11` is never defined in the installed `SDL_config_unix.h` header here:
https://github.com/libsdl-org/sdl2-compat/blob/876ce2b0e9a6c80f25568e42773f9ecfd61d9beb/include/SDL2/SDL_config_unix.h#L224

This is important because programs directly including `SDL_syswm.h` would fail to build on an X11-less system since `SDL_syswm.h` would try to include X11/Xlib headers on systems that don't have them.

I've been trying to figure out how to get this done. It seems to be related to `include/SDL2/SDL_config.h.in` which I don't remember to ever have touched directly before on the SDL2 days and only has some undefines for the X11 related stuff, so, please, how can I make it so `SDL2COMPAT_DISABLE_X11` is defined in `SDL_config_unix.h` when `-DSDL2COMPAT_DISABLE_X11=ON` is passed?

Thanks!

madebr:
SDL2's header were arch/build-specific: `include/SDL2/SDL_config.h` is generated (and installed) by cmake.
This does not happen anymore with SDL3 (which makes the headers fully generic). 
sdl2-compat's headers are also static, which I think is a plus.

#536 addresses the question posed here: `-DSDL2COMPAT_DISABLE_X11` will be added by the installed `sdl2-config`, `sdl2-compat.pc` and also when using `find_package(SDL2)`.

vanfanel:
@madebr 

It seems that building SDL2 programs that directly include `SDL_syswm.h` is still a problem:

```
/usr/local/include/SDL2/SDL_syswm.h:68:10: fatal error: X11/Xlib.h: No such file or directory
   68 | #include <X11/Xlib.h>
      |          ^~~~~~~~~~~~
```

Indeed, sdl2-config and pkg-config both pass `-DSDL2COMPAT_DISABLE_X11` as intended by your commit:
```
manuel@debian:~/src/hypseus-singe-2.11.6/b4$ sdl2-config --cflags
[...]

PR #566, opened 2025-12-28
Add SDL2COMPAT_DISABLE_X11 guards

vanfanel:
This is so programs that include SDL_syswm.h can build when X11 is disabled.

See: https://github.com/libsdl-org/sdl2-compat/issues/533#issuecomment-3694436279


slouken:
Merged, thanks!

PR #567, opened 2025-12-29
Generate include/SDL2/SDL_config_unix.h at build-time for SDL2COMPAT_DISABLE_X11

madebr:
By embedding `SDL2COMPAT_DISABLE_X11` in the header, there's no need for passing this macro through cmake or pkg-config.

For #533


slouken:
> By embedding `SDL2COMPAT_DISABLE_X11` in the header, there's no need for passing this macro through cmake or pkg-config.

Yep, good point. Let's remove that as part of this PR.

madebr:
> Yep, good point. Let's remove that as part of this PR.

This is done by removing the `target_compile_definitions(SDL2 PUBLIC SDL2COMPAT_DISABLE_X11)` line: CMake processes the [`INTERFACE_COMPILE_DEFINITIONS`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_COMPILE_DEFINITIONS.html) property for `sdl2-compat.pc`.

slouken (approved):

PR #572, opened 2026-01-05
Filter out non-fullscreen flags when setting fullscreen

### LINKED ISSUE #571: Descent3: game won't run fullscreen
version: 1.5.0 from Aug 16, 2024
sdl2-compat version: 2.32.62+ds-1from Debian experimental

I know this may seems low prio as recent git snapshot use SDL3, but the only released version is still using SDL2.

Steps to reproduce:
- cleanup everything: `rm -rf .local/share/Outrage\ Entertainment/Descent\ 3/`
- run the `Descent3` binary, the main menu is displayed fullscreen
- Start a new game. After the loading screen (still displayed fullscreen) you will ends up in your cockpit ready to fight. But the game is now rendered in a window and not fullscreen.
- If you press 'F2' and change the display resolution, the game goes back to fullscreen.
- Now press 'Esc' and abort the mission. You will be back to the main menu, but in a window and not fullscreen...

-> Each time you switch between the menu & the game, the screen switch to windowed mode.
This doesn't happen with classic SDL2.

Extra info: the problem occurs with both the X11 and the Wayland video driver.

Kontrabant:
Extra window flags being passed to SDL_SetWindowFullscreen will cause exact-matching to fail, so filter them out.

Fixes #571

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -9340,6 +9340,7 @@ SDL_SetWindowFullscreen(SDL_Window *window, Uint32 flags)
     int ret = 0;
     bool fullscreen = false;
     SDL_WindowFlags old_flags = SDL_GetWindowFlags(window);
+    flags &= SDL2_WINDOW_FULLSCREEN_DESKTOP;
 
     if (flags == SDL2_WINDOW_FULLSCREEN_DESKTOP) {
         fullscreen = true;


slouken:
Oh, gotcha, thanks.

src/sdl2_compat.c
```diff
@@ -9340,6 +9340,7 @@ SDL_SetWindowFullscreen(SDL_Window *window, Uint32 flags)
     int ret = 0;
     bool fullscreen = false;
     SDL_WindowFlags old_flags = SDL_GetWindowFlags(window);
+    flags &= SDL2_WINDOW_FULLSCREEN_DESKTOP;
```
smcv: The 1:1 equivalent of "classic" SDL2 seems to be

```suggestion
    flags &= SDL2_WINDOW_FULLSCREEN_DESKTOP | SDL2_WINDOW_FULLSCREEN;
```

PR #573, opened 2026-01-17
window: Set grab window flags even if grab fails

cgutman:
SDL3 made keyboard/mouse grab failable (in which case the window flags are not set), but SDL2 apps expect the old behavior where `SDL_WINDOW_KEYBOARD_GRABBED` and `SDL_WINDOW_MOUSE_GRABBED` are set regardless of whether the grab was successful.

Initially reported in https://github.com/moonlight-stream/moonlight-qt/issues/1776


slouken:
Merged, thanks!

PR #589, opened 2026-04-04
Expose SDL3 version string in SDL3_VERSION hint

### LINKED ISSUE #588: Querying the underlying SDL3 version
For apps that run on native SDL2, they can use `SDL_GetVersion()` to determine which SDL2 version they're using in case they need to apply specific workarounds for bugs in certain versions. When running under SDL2-compat, they can do the same to determine if workarounds are required for bugs in SDL2-compat itself, but there's no way to determine which SDL3 version is running under the hood. If there's an SDL3 bug that the app trips over, they don't have any way to know when to apply that workaround.

Possible options:
- Populate a hint for it when we load SDL3 (ex: `SDL3_VERSION` or something)
- Shove the SDL3 version in `SDL_GetRevisionNumber()` or similar (perhaps only with a specific hint set)
- Something else?
- Do nothing (assume that any maintained SDL2 app will eventually be ported to native SDL3)

My motivation for this is wanting to set `SDL_KMSDRM_ATOMIC=0` for SDL 3.4.0-3.4.2 due to https://github.com/libsdl-org/SDL/issues/15242 and https://github.com/moonlight-stream/moonlight-qt/issues/1824#issuecomment-4181496902

cgutman:
Fixes #588

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -1086,6 +1086,7 @@ LoadSDL3(void)
                         SDL2Compat_LogAtStartup(debugmsg);
                     }
                     SDL2Compat_ApplyQuirks(force_x11);  /* Apply and maybe print a list of any enabled quirks. */
+                    SDL2Compat_SetEnvAtStartup("SDL3_VERSION", sdl3verstr);
                 }
             }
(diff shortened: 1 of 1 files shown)


slouken:
Looks good!

Issue #590, opened 2026-04-14
Middle-earth: Shadow of Mordor: no gamepad rumble


RevengeRip:
When substituting shipped SDL2 library with sdl2-compat gamepad rumble doesn't work anymore.

RevengeRip:
Same happens in DiRT 4. I think all Feral Interactive games that had rumble support are now missing it with `sdl2-compat`.

cgutman:
I checked a few Feral Interactive titles, and it looks like those are using SDL haptics (specifically `SDL_HapticOpenFromJoystick()`) rather than the `SDL_GameControllerRumble()` API. In the case of DiRT 4, Shadow of Mordor, and Rise of the Tomb Raider, they _couldn't_ have used the modern rumble API because those titles shipped with SDL 2.0.7-2.0.8 while `SDL_GameControllerRumble()` didn't land until SDL 2.0.9. Shadow of the Tomb Raider ships with SDL 2.0.10, but I didn't check to see if they switched to `SDL_GameControllerRumble()`.

The removal of SDL haptics support for simple game controller rumble effects in SDL3 was intentional, but we may need to backtrack on that (or implement emulation in sdl2-compat). It would probably be cleaner to implement in SDL3 where we have full access to the haptic subsystem, but we could also do it in sdl2-compat by shimming all the haptics APIs and creating our own emulated haptic devices internally backed by `SDL_GameControllerRumble()`.

@slouken Thoughts?

slouken:
I think you're right. I'm not sure what the best approach is, but I'm reluctant to re-add controllers as haptic devices. Thoughts?

cgutman:
I started working on an SDL3 haptic emulation implementation and while I was debugging, I actually found a possible easy win for the near term. The Linux haptic implementation already has full support for joystick haptics on all joysticks where the kernel provides native FF support (most of them), we just explicitly block this with a `!SDL_IsGamepad()` check [here](https://github.com/libsdl-org/SDL/blob/41f079491a0e79b22441fd32a7c8ad91db237744/src/haptic/SDL_haptic.c#L311). Simply removing that check restores haptic support for Xbox and PlayStation controllers in Feral Interactive titles.

Obviously this isn't a full solution (doesn't work with HIDAPI, Windows, iOS, Android, or most other platforms), but it's much better than nothing and is probably simple enough to ship in 3.4.6 while we work on another solution. What do you think?

slouken:
[...]

PR #594, opened 2026-05-01
hint: Add SDL2_COMPAT hint for SDL3 compatibility features

cgutman:
This enables the compatibility behavior for rumble support being introduced in https://github.com/libsdl-org/SDL/pull/15488 to partially address https://github.com/libsdl-org/sdl2-compat/issues/590.

I opted to use `SDL2Compat_SetEnvAtStartup()` instead of `SDL3_SetHint()` to avoid the hint getting nuked if the app calls `SDL_ResetHints()`.


slouken:
Merged, thanks!

Issue #340, opened 2025-02-10
MinGW build of 2.32.50 fails on Fedora with a link error


Conan-Kudo:
When attempting to build v2.32.50 on Fedora for MinGW, I get the following error:

```
[ 12%] Linking C shared library SDL2.dll
/usr/bin/cmake -E cmake_link_script CMakeFiles/SDL2.dir/link.txt --verbose=1
/usr/bin/cmake -E rm -f CMakeFiles/SDL2.dir/objects.a
/usr/bin/i686-w64-mingw32-ar qc CMakeFiles/SDL2.dir/objects.a @CMakeFiles/SDL2.dir/objects1.rsp
/usr/bin/i686-w64-mingw32-gcc -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions --param=ssp-buffer-size=4  -Wall -Wdeclaration-after-statement -DSDL_HAVE_WINAPIFAMILY_H  -nostdlib -fstack-protector -lssp -Wl,--no-undefined -shared -o SDL2.dll -Wl,--out-implib,libSDL2.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -Wl,--whole-archive CMakeFiles/SDL2.dir/objects.a -Wl,--no-whole-archive @CMakeFiles/SDL2.dir/linkLibs.rsp
make[2]: Leaving directory '/builddir/build/BUILD/mingw-sdl2-compat-2.32.50-build/sdl2-compat-release-2.32.50/build_win32'
/usr/lib/gcc/i686-w64-mingw32/14.2.1/../../../../i686-w64-mingw32/bin/ld: CMakeFiles/SDL2.dir/objects.a(sdl2_compat.c.obj): in function `SDL2Compat_strlen':
/builddir/build/BUILD/mingw-sdl2-compat-2.32.50-build/sdl2-compat-release-2.32.50/src/sdl2_compat.c:255:(.text+0x3312): undefined reference to `strlen'
collect2: error: ld returned 1 exit status
```

Full build log: [mingw-sdl2-compat-2.32.50-build.log.txt](https://github.com/user-attachments/files/18731640/mingw-sdl2-compat-2.32.50-build.log.txt)

I'm pretty sure this was caused by some of the refactoring for the platform headers, but I'm not sure how exactly.

sezero:
[...]

PR #603, opened 2026-06-13
tweak SDL2Compat_strlen() / volatile hack to be c++-23 compatible.

sezero:
Reference issue: https://github.com/libsdl-org/sdl2-compat/issues/340.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -272,11 +272,11 @@ SDL2COMPAT_itoa(char *dst, int val)
 /* you can use SDL3_strlen once we're past startup. */
 static int SDL2Compat_strlen(const char *str)
 {
-    volatile int retval = 0;  /* volatile prevents gcc from optimizing this into strlen() */
-    while (str[retval]) {
-        retval++;
+    volatile const char *ptr = str;
+    while (*ptr) {
+        ++ptr;
     }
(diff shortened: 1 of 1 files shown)


src/sdl2_compat.c
```diff
@@ -272,11 +272,11 @@ SDL2COMPAT_itoa(char *dst, int val)
 /* you can use SDL3_strlen once we're past startup. */
 static int SDL2Compat_strlen(const char *str)
 {
-    volatile int retval = 0;  /* volatile prevents gcc from optimizing this into strlen() */
-    while (str[retval]) {
-        retval++;
```
madebr, line 277: The error was: `'++' expression of 'volatile'-qualified type is deprecated`
Would this also work?
```
retval = retval + 1;
```

sezero: _Maybe_, I don't know (have to reboot to linux to test, can't now.) You can test by g++-16 or g++15 with -std=gnu++23 to see.

Is there anything wrong with the new code, btw?

madebr (approved):

src/sdl2_compat.c
madebr: I just tested on my machine, and it would also work.
But your code is fine as well.

PR #605, opened 2026-06-15
Use Standard C strlen(), etc., if we're using libc anyway

smcv:
As I suggested on #340 a while ago.

* Restore comment explaining why volatile is used in SDL2Compat_strlen()

* Make SDL2Compat_strlen return size_t
    
    This gives it the same signature as Standard C strlen(), and avoids the
    possibility of signed integer overflow.

* SDL2Compat_ApplyQuirks: Avoid overflow for extremely long exe_name
    
    The length of `exe_name` is (at least in theory) unbounded, so adding
    anything to it could overflow, but we know that the right-hand side
    of the `<=` evaluates to a positive constant.

* Use Standard C strlen(), etc., if we're using libc anyway
    
    On some platforms (notably Windows) we don't want to use the system C
    library, so we avoid using Standard C strlen() and use our own
    implementation. On other platforms (notably Linux) we require the system
    C library anyway, for functions like setenv(), so we might as well make
    full use of it.
    
    In particular this means that our fallback implementations only need
    to work on Windows, so it's now OK if the workarounds we use to avoid the
    compiler replacing our open-coded strlen() with a call to the library
    strlen() (see #340) are not 100% portable.

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -272,6 +272,7 @@ SDL2COMPAT_itoa(char *dst, int val)
 /* you can use SDL3_strlen once we're past startup. */
 static size_t SDL2Compat_strlen(const char *str)
 {
+#ifdef SDL_PLATFORM_WINDOWS
     /* volatile prevents gcc from optimizing this into a call to the
      * strlen() library function, which we are intentionally avoiding on
      * Windows: see #340 */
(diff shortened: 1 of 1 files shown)


smcv:
~~Hmm, this isn't quite right: on native Linux I'm still ending up in the `!HAVE_LIBC` code path.~~ — Fixed

smcv:
@Conan-Kudo, @sezero: Please could you try this in the build configuration where #340 was reproduced? It should take the open-coded code path there.

Verified to take the libc `strlen()` code path in a native build on Linux (I added a `#error` to the open-coded version to make sure it wasn't chosen) and passes automated tests. Otherwise untested.

sezero (approved):
Build tested OK for i686-linux, and i686- and x86_64-mingw32.

slouken (approved):

icculus (approved):

slouken:
Merged, thanks!

PR #607, opened 2026-06-18
Scale pointer coordinates even when logical presentation is disabled

### LINKED ISSUE #606: Caesar 3 (Augustus) works with SDL2 but doesn't with SDL2-compat if the screen scaling option is enabled
Platform: Linux

[Augustus](https://github.com/Keriew/augustus) (the fork of [Julius](https://github.com/bvschaik/julius), and open source implementation of the classic game Caesar 3) has an option to increase it's screen scaling.
While using the regular SDL2 library the mouse clicks work as expected in any % of scaling.
While using SDL2-compat (like the one on fedora's repository), as soon as the scaling is set to anything other than 100%, the mouse clicks get messed up as they are offset by a set amount and if you want to click on a menu item for example, you have hover the mouse under and to the right of it until the button gets highlighted.
I made my tests by using the latest appimage build of the game, and extracting it using the --appimage-extract argument.
After that, i started compiling many SDL2 versions and linking to it via the LD_LIBRARY_PATH environment variable.
The game worked fine with any SDL2 .so file from 2.0.20 to the most recent, but bugged as soon as i used a .so from SDL2-compat.

Please feel free to ask me any questions about this

Kontrabant:
Logical presentation will be reported as disabled after setting a render target, however, pointer coordinates may still need to be scaled, as the previous view state is still preserved internally (the 'main_view' in the SDL3 render backend).

Fixes #606

### DIFF
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -2304,22 +2304,18 @@ static SDL2_Event *Event3to2(const SDL_Event *event3, SDL2_Event *event2)
     case SDL_EVENT_MOUSE_MOTION:
         renderer = SDL3_GetRenderer(SDL3_GetWindowFromID(event3->motion.windowID));
         if (renderer) {
-            SDL_RendererLogicalPresentation mode = SDL_LOGICAL_PRESENTATION_DISABLED;
-            SDL3_GetRenderLogicalPresentation(renderer, NULL, NULL, &mode);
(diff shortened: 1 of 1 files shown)


bjorn:
Unfortunately this caused a regression in Mana client, making it entirely unusable when scaling is enabled. :-( See #624 for more information.

PR #612, opened 2026-07-15
Add SDL3 dlopen note

madebr:
This bumps the minimum required SDL3 version at build-time to 3.4.

Actually, I don't know what is the minimum required SDL3 version.

```
$ dlopen-notes.py libSDL2-2.0.so.0
# libSDL2-2.0.so.0
[
  {
    "feature": "SDL3",
    "description": "sdl2-compat translates SDL2 API calls into equivalent SDL3 calls",
    "priority": "required",
    "soname": [
      "libSDL3.so.0"
    ]
  }
]
```


slouken:
I think the later the better, but I believe we only _require_ 3.4.0

tom--pollard:
Fwiw the minimum requirement used to be 3.2.0 before this, and should have probably been better communicated in the commit / release notes

slouken:
> Fwiw the minimum requirement used to be 3.2.0 before this, and should have probably been better communicated in the commit / release notes

In general, the latest sdl2-compat release should be paired with the latest SDL release.

PR #617, opened 2026-08-06
Cygwin/Mingw fix support requires patch to SDL 3.4.x

stahta01:
I have gotten SDL 3.4.x and sdl-compat repos to build with these cmake options:
```
    -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc
    -DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++
    -DCMAKE_RC_COMPILER=x86_64-w64-mingw32-windres
```
SDL 3.4.x patch needed
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 1785395fb..ed7461ec5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,13 @@ if(POLICY CMP0054)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
+if(CYGWIN AND CMAKE_C_COMPILER MATCHES ".*[-]mingw32[-].*")
+  set(MINGW 1)
+  set(WINDOWS 1)
+  set(WIN32 1)
+  set(UNIX 0)
+endif()
+
 include(CheckLibraryExists)
 include(CheckIncludeFile)
```

I have not yet tested the built binary except for doing the normal ctest tests; the sdl2-config looked okay, but it should be tested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,13 @@ set(SDL_MINOR_VERSION ${sdl2_compat_VERSION_MINOR})
 set(SDL_MICRO_VERSION ${sdl2_compat_VERSION_PATCH})
 set(SDL_VERSION "${SDL_MAJOR_VERSION}.${SDL_MINOR_VERSION}.${SDL_MICRO_VERSION}")
 
+if(CYGWIN AND CMAKE_C_COMPILER MATCHES ".*[-]mingw32[-].*")
+  set(MINGW 1)
+  set(WINDOWS 1)
+  set(WIN32 1)
+  set(UNIX 0)
+endif()
+
 include(CheckCSourceCompiles)
 include(CheckIncludeFile)
 include(CheckCCompilerFlag)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -388,7 +388,12 @@ else()
   )
 endif()
 
-if(MINGW)
+if(CYGWIN AND CMAKE_C_COMPILER MATCHES ".*[-]mingw32[-].*")
+  set_property(TARGET SDL2 APPEND_STRING PROPERTY LINK_FLAGS " -nostdlib")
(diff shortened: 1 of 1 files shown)


sezero:
I don't like this: You are violating the cmake'ry
And you _don't_ actually need cmake to build sdl2-compat in a cygwin
environment: Why don't you use the mingw makefile for it, e.g.:
```sh
make -C src -f Makefile.mingw CROSS=x86_64-w64-mingw32
```

The only patch that can be applied to sdl2-compat can be the following
(not tested), that is assuming that you want to use the cygwin version
of SDL3.dll:

```diff
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index c8d0eec..3bb71d5 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -385,7 +385,8 @@ static char loaderror[256];
     static HMODULE Loaded_SDL3 = NULL;
     #define DIRSEP "\\"
     #define SDL3_LIBNAME "SDL3.dll"
+    #define SDL3_LIBNAME2 "cygSDL3.dll"
-    #define LoadSDL3Library() ((Loaded_SDL3 = LoadLibraryA(SDL3_LIBNAME)) != NULL)
+    #define LoadSDL3Library() ((Loaded_SDL3 = LoadLibraryA(SDL3_LIBNAME)) != NULL || (Loaded_SDL3 = LoadLibraryA(SDL3_LIBNAME2)) != NULL)
     #define LookupSDL3Sym(sym) (void *)GetProcAddress(Loaded_SDL3, sym)
     #define CloseSDL3Library() { if (Loaded_SDL3) { FreeLibrary(Loaded_SDL3); Loaded_SDL3 = NULL; } }
 #elif defined(__APPLE__)
```

stahta01:
Too many different people saying different things this is no longer worth the effort.

Issue #621, opened 2026-08-19
UQM: main menu highlight doesn't pulse and renders black under sdl2-compat


Amon-X:
### Description

In the main menu, the selected menu item normally pulses/flashes to indicate selection. Under `sdl2-compat`, this animation doesn't happen. The selected item is static. The in-game menu also renders black instead of drawing correctly.

Neither issue occurs with native `libsdl2-classic`.

### Environment

* OS: Debian forky/sid x86_64, kernel 7.1.7+deb14-amd64
* Display: X11, windowed
* DE: Xfce 4.20
* sdl2-compat: 2.32.70 (built Jun 8 2026)
* SDL3: 3.4.14
* UQM: v0.8.0 (compiled Oct 29 2025)

### Reproduce

1. Launch UQM windowed under X11 via `sdl2-compat`
2. Open the main menu / Crew or Fuel purchase screens
3. Observe the selected menu item (no pulse) and menu background (renders black)

### Control test

Native `libsdl2-classic`: works fine. Regression is specific to `sdl2-compat`. Game displays correctly via 
SDL_DYNAMIC_API='/usr/$LIB/sdl2-classic/libSDL2-2.0.so.0' uqm

### Logs

```
sdl2-compat 2.32.70, built on Jun  8 2026 at 16:55:46, talking to SDL3 3.4.14
sdl2-compat: This app appears to be named: uqm
The Ur-Quan Masters v0.8.0 (compiled Oct 29 2025 12:59:57)
Netplay protocol version 0.4. Netplay opponent must have UQM 0.6.9 or later.
SDL_DYNAMIC_API: Not overridden
Using SDL version 2.32.70 (compiled with 2.32.10)
```

### Suspected root cause

~~UQM's SDL2 backend renders into a software surface and only pushes the changed rectangle to a GPU streaming texture via `SDL_UpdateTexture()` each frame. `TFB_SDL2_Preprocess()` decides whether that pushed rect is the full screen or just the dirty bounding box:~~

### Actual root cause
 From one of the devs : McMartin:  "The issue here is that UQM's code relied on SDL_GetColorKey to not touch the output variable when the call fails. The versions of SDL2 we'd used in the past left it alone in such cases; SDL3 appears to force the value to zero. That broke the transparency checks. "

### Expected vs Actual

Expected: selected menu item pulses/flashes; menu renders normally (`expected.gif`). 
[...]

PR #623, opened 2026-08-22
More faithfully track SDL_GetColorKey semantics

michaelcmartin:
SDL3_GetSurfaceColorKey starts by writing 0 to *key if non-NULL. (see: https://github.com/libsdl-org/SDL/blob/main/src/video/SDL_surface.c#L703)

SDL_GetColorKey only does so if the call will ultimately succeed. (see: https://github.com/libsdl-org/SDL/blob/release-2.32.x/src/video/SDL_surface.c#L392)

This patch updates the sdl2-compat implementation to leaves the value of *key untouched on failure paths. This fixes issue #621 in UQM 0.8 where the flashrect does not function properly.


slouken:
Merged, thanks!

PR #625, opened 2026-09-17
Scale mouse coordinates only when the main view has a logical presentation mode set

### LINKED ISSUE #624: Mouse event coordinates are scaled twice when only SDL_RenderSetScale is used (2.32.72)
Since 2.32.72, mouse motion and button events arrive already divided by the render scale when an application uses `SDL_RenderSetScale` without ever calling `SDL_RenderSetLogicalSize`. In 2.32.70 and in real SDL2 the events are in window coordinates in that case.

SDL2's `SDL_RendererEventWatch` only rescales events when a logical size is set ([src/render/SDL_render.c on the SDL2 branch](https://github.com/libsdl-org/SDL/blob/b90ac95029d801c5abc59472ba8e2200dff31e1e/src/render/SDL_render.c#L774-L785)):

```c
    } else if (event->type == SDL_MOUSEMOTION) {
        SDL_Window *window = SDL_GetWindowFromID(event->motion.windowID);
        if (window == renderer->window) {
            int logical_w, logical_h;
            SDL_DRect viewport;
            SDL_FPoint scale;
            GetWindowViewportValues(renderer, &logical_w, &logical_h, &viewport, &scale);
            if (logical_w) {
                event->motion.x -= (int)(viewport.x * renderer->dpi_scale.x);
                event->motion.y -= (int)(viewport.y * renderer->dpi_scale.y);
                event->motion.x = (int)(event->motion.x / (scale.x * renderer->dpi_scale.x));
                event->motion.y = (int)(event->motion.y / (scale.y * renderer->dpi_scale.y));
```

This looks like a side effect of cf62e9a ("Scale pointer coordinates even when logical presentation is disabled"). Before that commit the conversion was guarded by the presentation mode:

https://github.com/libsdl-org/sdl2-compat/blob/991bed47d4d4cb0485a7c9f3d9d4f25281be5740/src/sdl2_compat.c#L2322-L2329

After it the events are always converted:

https://github.com/libsdl-org/sdl2-compat/blob/cf62e9ad153f4cac3c5e1a782ba357c2e5bd350c/src/sdl2_compat.c#L2322-L2326

The same change was made for the `SDL_EVENT_MOUSE_BUTTON_DOWN`/`SDL_EVENT_MOUSE_BUTTON_UP` and `SDL_EVENT_MOUSE_WHEEL` cases. `SDL3_ConvertEventToRenderCoordinates` applies the render scale regardless of the presentation mode, so an application that never asked for logical presentation now gets scaled events.

The Mana client (https://git.themanaworld.org/mana/mana) sets a render scale of 2 for high resolution displays and converts each event with `SDL_RenderWindowToLogical`. With 2.32.72 that conversion is applied on top of the already scaled event, so every click lands at half the expected distance from the top left corner and the GUI becomes unusable at scale 2. `SDL_GetMouseState` still reports window coordinates, so the two disagree by the scale factor.

Reproduce: create a window and renderer, call `SDL_RenderSetScale(renderer, 2.0f, 2.0f)` and no logical size, then move the mouse to window position (200, 100). `SDL_GetMouseState` reports (200, 100) while the `SDL_MOUSEMOTION` event reports (100, 50). With 2.32.70 both report (200, 100).

Seen on Fedora 44 with sdl2-compat 2.32.72 and SDL3 3.4.16.

Kontrabant:
- [X] I confirm that I am the author of this code and release it to the SDL project under the zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

When a render target is set, querying the logical presentation mode will report the presentation mode on the render target, which can differ from the main view, and the main view properties are what is used to transform mouse coordinates. When a render target is bound, store the main view presentation mode, and use it to determine whether mouse coordinates in events should be implicitly scaled.

This should be a proper fix for the issue.

Fixes #624 without regressing #606


Kontrabant:
Looking more at SDL2, this is definitely the correct fix for the issue, as SDL2 doesn't implicitly scale event values unless the window itself is using a logical presentation mode, and when a render target is attached, the 'backup' logical values are queried, which are equivalent to the "main view" in SDL3.

bjorn:
Thanks for the quick fix!

Kontrabant:
No problem. Sorry about letting this bug slip through in the first place :man_facepalming:.

PR #626, opened 2026-09-23
Support latest QNX driver.

eleir9268:
- [x] I confirm that I am the author of this code and release it to the SDL project under the zlib license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

## Description

This adds support for the latest changes to QNX video in SDL3.

I tested this exact version by running the unit tests and Maelstrom, but these changes were used under the hoods of OGRE and MAME as well.


src/sdl2_compat.c
```diff
@@ -4825,6 +4829,10 @@ SDL_DECLSPEC SDL2_bool SDLCALL SDL_GetWindowWMInfo(SDL_Window *window, SDL2_SysW
         info->subsystem = SDL2_SYSWM_ANDROID;
         info->info.android.window = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_ANDROID_WINDOW_POINTER, NULL);
         info->info.android.surface = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_ANDROID_SURFACE_POINTER, NULL);
+    } else if (SDL3_strcmp(driver, "qnx") == 0) {
+        info->subsystem = SDL2_SYSWM_QNX;
+        info->info.android.window = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_QNX_WINDOW_POINTER, NULL);
+        info->info.android.surface = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_QNX_SURFACE_POINTER, NULL);
```
slouken:
```suggestion
        info->info.qnx.window = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_QNX_WINDOW_POINTER, NULL);
        info->info.qnx.surface = SDL3_GetPointerProperty(props, SDL_PROP_WINDOW_QNX_SURFACE_POINTER, NULL);
```

eleir9268: Fixed, thanks!

src/sdl2_compat.c
```diff
@@ -100,6 +100,10 @@ This breaks the build when creating SDL_ ## DisableScreenSaver
 #include <unistd.h> /* for readlink() */
 #endif
 
+#ifdef __QNX__
```
slouken: Is there a reason you're using `__QNX__` instead of `SDL_PLATFORM_QNXNTO`?

eleir9268: I wrote this a while ago, but I was copying the adjacent formatting. I will swap it over to using the platform macro instead.

src/sdl2_compat.h
```diff
@@ -1300,6 +1300,7 @@ typedef enum
   SDL2_SYSWM_MIR,  /* no longer available, left for API/ABI compatibility. */
   SDL2_SYSWM_WINRT,
   SDL2_SYSWM_ANDROID,
+  SDL2_SYSWM_QNX,
```
slouken: You can't add a value to the middle of the enum, this breaks binary compatibility. This isn't in the public SDL2 API. Are you shipping new SDL2 headers with this that include this change for QNX? If not, nobody can use this change and we should drop it.

eleir9268: Makes sense, thanks. I dropped it. To my knowledge we're not relying on this for anything.

slouken:
Merged, thanks!

